PR #590, opened 2015-01-29
588

### LINKED ISSUE #588: biom validate-table requires h5py
`biom validate-table` requires that h5py/HDF5 are installed, even when validating a JSON-based file:

```
IOError: h5py is not installed, can only validate JSON tables
```

Jorge-C:
Fix #588.

It is slightly hackish because pyqi will say that `is_json == False` without really knowing whether it is actually a json file or not, we must parse it and decide whether it looks json-ish or hdf5-ish, and ignore pyqi's opinion :(

### DIFF
--- a/biom/commands/table_validator.py
+++ b/biom/commands/table_validator.py
@@ -72,7 +72,11 @@ class TableValidator(Command):
     HDF5FormatVersions = set([(2, 0), (2, 0, 0), (2, 1), (2, 1, 0)])
 
     def run(self, **kwargs):
-        is_json = kwargs['is_json']
+        # We can't trust kwargs['is_json'] because that's determined
+        # before any parsing happens...
+        with open(kwargs['table']) as f:
+            # from the HDF5 documentation about format signature
+            is_json = f.read(8) != '\x89HDF\r\n\x1a\n'
 
         if kwargs['format_version'] in [None, 'None']:
             if is_json:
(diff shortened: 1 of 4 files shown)


biom/commands/table_validator.py
```diff
@@ -72,7 +72,11 @@ class TableValidator(Command):
     HDF5FormatVersions = set([(2, 0), (2, 0, 0), (2, 1), (2, 1, 0)])
 
     def run(self, **kwargs):
-        is_json = kwargs['is_json']
+        # We can't trust kwargs['is_json'] because that's determined
+        # before any parsing happens...
+        with open(kwargs['table']) as f:
```
ElDeveloper: We are doing the same in `biom_open` so I'm inclined to say that we should move these two lines into a function living in `util.py`, maybe `is_hdf5_file`? Depending on what you think, a test should also be added.

ElDeveloper:
Thanks for working on this, I am not super familiar with `pyqi`, but maybe @jairideout or @ebolyen can have a look and make sure this is ok.

Jorge-C:
Yeah, I'm not really familiar with it either, so maybe there's another way to achieve that.

:+1: on a util function and test. Will work on that later today.

jairideout:
What is the purpose of `--is-json`? It seems silly to have to tell the script what the underlying file format is of a file that you're trying to validate (most users won't know what this is). Perhaps this option could be removed?

Jorge-C:
Addressed @ElDeveloper's comments, and created issue #591 to keep track of @jairideout's suggestion.

This is ready!

ElDeveloper:
Thanks @Jorge-C!

PR #596, opened 2015-02-05
Fix595

### LINKED ISSUE #595: BIOM convert is not recording version information correctly
If you convert an HDF5 table (format 2.1.0) to JSON, the resulting file format indicates it is 2.1.0.

```bash
(python27)12:29:55 examples@master$ biom convert -i min_sparse_otu_table_hdf5.biom -o foo --to-json
(python27)12:30:15 examples@master$ cut -c 1-80 foo
{"id": "No Table ID","format": "Biological Observation Matrix 2.1.0","format_url
```

This is incorrect as the JSON file format is actually version 1.0.0.

Jorge-C:
Any ideas on how to add a test for this?

Fixes #595.

### DIFF
--- a/biom/commands/table_converter.py
+++ b/biom/commands/table_converter.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-
+# -*- coding: utf-8 -*-
 # ----------------------------------------------------------------------------
 # Copyright (c) 2011-2013, The BIOM Format Development Team.
 #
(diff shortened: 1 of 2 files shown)


wasade:
Not worth it imo, mergable?

adamrp:
I don't think it requires a test, either, but looking at the code does raise one question. Do we want to record the code version as well as the format version? That way, we know this is biom *format* 1.0 that was written with X version of the code (might make it easier to track bugs in the future).

wasade:
We'd need to redefine both BIOM-Format 1.0 and 2.0

Jorge-C:
Then it's already mergable.

PR #602, opened 2015-03-03
Pep8 updates

wasade:
This is blocking other pending PRs. Should resolve outstanding pep8 issues.

### DIFF
--- a/biom/util.py
+++ b/biom/util.py
@@ -40,6 +40,8 @@ __license__ = "BSD"
 __url__ = "http://biom-format.org"
 __maintainer__ = "Daniel McDonald"
 __email__ = "daniel.mcdonald@colorado.edu"
+__format_version__ = (2, 1)
+__version__ = "2.1.3-dev"
 
 
 def generate_subsamples(table, n, axis='sample', by_id=False):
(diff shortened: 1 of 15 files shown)


josenavas:
:+1:

biom/util.py
```diff
@@ -40,6 +40,8 @@
 __url__ = "http://biom-format.org"
 __maintainer__ = "Daniel McDonald"
 __email__ = "daniel.mcdonald@colorado.edu"
+__format_version__ = (2, 1)
```
ElDeveloper: Why were these two things moved here?

ElDeveloper:
*nooooooooo* ... oh well :L

josenavas:
@ElDeveloper why don't you gitter?

wasade:
@ElDeveloper, its the classic chicken and pep8 issue, where the module
level imports need to be the first thing. Normally this is fine, except
that biom.table and biom.parse depends on the versions, so the module level
import of those results in a cyclic hell. This was the only solution I
could find that would preserve API compatibility (for instance qiime's
print_qiime_config.py does a `from biom import __version__`)

ElDeveloper:
OK, that was exactly my concern. So then it should work just fine right?
i.e. projects that do `from biom import __version__` will continue to
work without a problem?

wasade:
That is the intent

ElDeveloper:
:boom: *slow :clap:*

PR #603, opened 2015-03-04
591

Jorge-C:
We were already ignoring `--is-json`, now remove it. Note that this is not backwards compatible, as trying to validate a table using that option will raise an error: `biom validate-table: error: no such option: --is-json`.


wasade:
@jairideout @gregcaporaso, this PR will introduce a change to the command line usage. Please advise with respect to the minor release

jairideout:
I think this change is okay for QIIME because QIIME isn't using this script (grepping for `validate-table` didn't have any hits).

Jorge-C:
Thanks @jairideout!

wasade:
im :+1:

PR #604, opened 2015-03-05
DOC: listing 2.1.4 changes

wasade:
One step closer to 2.1.4. We need to deploy and test against QIIME before the release is actually cut, marking as 2.1.4rc1 for now.

NOTE: version tags across the codebase have _not_ been updated. Will do that on release.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,8 +1,8 @@
 BIOM-Format ChangeLog
 =====================
 
-biom 2.1.3rc1
--------------
+biom 2.1.4
+----------
 
 Changes:
 


Jorge-C:
Something odd is going on with the newest cython version (I can reproduce locally)

gregcaporaso:
To test for QIIME, easiest might be to just boot up the latest VirtualBox or AWS instance, install this version of biom via ``sudo pip install http://....zip --upgrade`` and then run ``all_tests.py``.

wasade:
@gregcaporaso, running on pando right now in essentially that though the install isn't a complete one. Good call on VB, will pull and boot

@Jorge-C, that is weird. Looks like cython got a little more picky over types that should be compatible AFAIK. Do you think you can push a fix in on it?

Jorge-C:
Yes, I'm working on that right now!

wasade:
Great, thanks!

Just remembered I'm on a home connection right now, so the 3.6GB download for the VM sucks, and I don't have any AWS creds at the moment. Will pull image once I get on campus

gregcaporaso:
Oh... if you're pretty confident that this will work with QIIME, the
easiest is to just do a biom release, update the minimum dependency in
QIIME's setup.py, and issue a pull request. Jenkins will then pull that
version of biom and use it for testing.

Do I have all of that right @jairideout?

jairideout:
That sounds right. Do we want to update the minimum required version in QIIME or keep it as-is? We didn't update the minimum required version for scikit-bio when we pushed a minor bug fix release.

Either way, issuing a PR like this will kick off the tests and allow us to see if there's any issues (we can always close it afterwards).

gregcaporaso:
I was thinking we could just leave the same minimum version number, but just use this to kick off the tests. Though I guess updating the minimum in QIIME 1.9.0-dev doesn't hurt...

wasade:
The downside is that we potentially cut a release that we then need to
backtrack, unless I'm misunderstanding?

jairideout:
Good point, I guess we might as well update the minimum version for 1.9.0-dev (could do this for scikit-bio too).

gregcaporaso:
@wasade, I don't understand your last question, sorry. Can you clarify?

wasade:
@gregcaporaso, you're recommending that we just do the release and then test against qiime, right? Doesn't that expose us to the possibility of needing to retract the release if there is some aspect of 2.1.4 that is not qiime 1.9 compatible?

gregcaporaso:
Got it. Definitely, I think it'd only be a good way to go if you're 99% sure that it works with QIIME, and just want to do a last test to confirm.

wasade:
Will wait to pull down the VM and test directly :)

Jorge-C:
Build is now green, this can be merged, right?

wasade:
Unsure... I've got some test failures but many are due to an incomplete
install. investigating

wasade:
Ya, I think the failures are all due to an incomplete install of qiime in
my environment. Need to test on a full install, which will be easiest on a
VM. I'll be heading to better internet somewhat soon

wasade:
@gregcaporaso, shouldn't numpy be available immediately on the VM?
screen shot 2015-03-05 at 3 53 16 pm

wasade:
...I tried executing `/qiime_software/activate.sh` but still no numpy...

ElDeveloper:
I just tried in my VM and it works:

screen shot 2015-03-05 at 2 56 52 pm

wasade:
I've got a fresh copy of 1.9 VM...
screen shot 2015-03-05 at 3 59 59 pm

wasade:
Did coveralls change its reporting? I'm getting a -7% coverage but the report isn't showing any change.

wasade:
somehow i hosed a fresh env... no idea how

wasade:
Okay, got these. @gregcaporaso @jairideout, are these expected? No usearch in my environment. 
screen shot 2015-03-05 at 5 44 25 pm

jairideout:
Those look like expected failures, @gregcaporaso do you agree? Probably the safest thing to do is scroll through the output and make sure the errors are only related to missing installs of usearch, torque, and sfffile/sffinfo.

gregcaporaso:
That looks good, those are the usual failures from the missing apps that @jairideout mentioned. I do usually scroll through the full output just to confirm (we clearly need to be using conditional unit testing there... QIIME 2...).

wasade:
I believe this is good to merge and release

wasade:
Just an update on this, I'm still working through some issues for the biom 2.x update to picrust and it is possible this will trigger a few other changes to biom. I'd like to hold off release of 2.1.4 until the picrust is good to go

PR #605, opened 2015-03-05
Cython

Jorge-C:
Update extension sources, and fix compatibility issue with Cython 0.22


wasade:
That's annoying... thanks @Jorge-C!

PR #607, opened 2015-03-07
Issue 606

### LINKED ISSUE #606: Table.collapse in one_to_many was not producing the right data shape
This issue wasn't previously identified as it is likely that only picrust uses this path through the code. The logic changed some when we merged the collapse methods from 1.3.1 to 2.x.

wasade:
Fixes #606

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -2233,15 +2233,17 @@ class SparseTableTests(TestCase):
 
     def test_collapse_observations_by_metadata_one_to_many(self):
         """Collapse observations by arbitary metadata"""
-        dt_rich = Table(np.array([[5, 6, 7], [8, 9, 10], [11, 12, 13]]),
-                        ['1', '2', '3'], ['a', 'b', 'c'],
+        dt_rich = Table(np.array([[5, 6, 7], [8, 9, 10], [11, 12, 13],
+                                  [14, 15, 16]]),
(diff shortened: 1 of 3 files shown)


tests/test_table.py
```diff
@@ -2233,15 +2233,17 @@ def bin_f(id_, x):
 
     def test_collapse_observations_by_metadata_one_to_many(self):
         """Collapse observations by arbitary metadata"""
-        dt_rich = Table(np.array([[5, 6, 7], [8, 9, 10], [11, 12, 13]]),
-                        ['1', '2', '3'], ['a', 'b', 'c'],
+        dt_rich = Table(np.array([[5, 6, 7], [8, 9, 10], [11, 12, 13],
```
ElDeveloper: Instead of modifying the existing case, can you leave this one and add a new one to reproduce the bug?

ElDeveloper:
Just one minor comment, can you also add a note to the ChangeLog?

Jorge-C:
I'm afraid I don't understand the collapse function to comment on it, but the fix doesn't look weird ;)

wasade:
This is an improvement on the existing test

ElDeveloper:
I thought you said the test case did not catch this because the inputed table was squared, I wonder if we can have a case where the table is squared and one where it isn't.

wasade:
There still is

ElDeveloper:
Now I got it! Thanks :+1:

PR #610, opened 2015-03-07
ENH/TST: custom parsers and formatters for hdf5 metadata

### LINKED ISSUE #608: Table.to_hdf5 was not formatting COG_Category correctly
The formatters for the metadata are [hardcoded](https://github.com/biocore/biom-format/blob/master/biom/table.py#L3471) such that, if a key that corresponds to a hardcoded formatter is not found, it will default to the general formatter. This means that `vlen_list_of_str_formatter` is only accessible for the hardcoded types. This should be parameterized, however, that introduces an API change. 

The quick fix here is to allow a hardcoding for "COG_Category". The better fix requires expanding out the `to_hdf5` method resulting in an API change.

cc @gregcaporaso @jairideout regarding API change or quick fix. I'm leaning on quick fix...

wasade:
Fixes #608 

This change allows for custom parsers and formatters. I don't think that it is sustainable adding in more hardcoded parsing types. Within picrust, we'll be able to just define a parser/formatter that serializes the nested lists in a few of the metadata types into json.

### DIFF
--- a/biom/table.py
+++ b/biom/table.py
@@ -2975,7 +2975,7 @@ class Table(object):
                               sample_ids[:], obs_md, sample_md)
 
     @classmethod
-    def from_hdf5(cls, h5grp, ids=None, axis='sample'):
+    def from_hdf5(cls, h5grp, ids=None, axis='sample', parse_fs=None):
         """Parse an HDF5 formatted BIOM table
 
         If ids is provided, only the samples/observations listed in ids
@@ -3066,6 +3066,12 @@ dataset of int32
             to retrieve from the hdf5 biom table
         axis : {'sample', 'observation'}, optional
             The axis to subset on
+        parse_fs : dict, optional
+            Specify custom parsing functions for metadata fields. This dict is
+            expected to be {'metadata_field': function}, where the function
+            signature is (object) corresponding to a single row in the
+            associated metadata dataset. The return from this function an
+            object as well, and is the parsed representation of the metadata.
 
         Returns
         -------
@@ -3115,6 +3121,9 @@ html
         if axis not in ['sample', 'observation']:
             raise UnknownAxisError(axis)
 
+        if parse_fs is None:
+            parse_fs = {}
+
         id_ = h5grp.attrs['id']
         create_date = h5grp.attrs['creation-date']
         generated_by = h5grp.attrs['generated-by']
@@ -3140,6 +3149,7 @@ html
             parser['taxonomy'] = vlen_list_of_str_parser
             parser['KEGG_Pathways'] = vlen_list_of_str_parser
             parser['collapsed_ids'] = vlen_list_of_str_parser
+            parser.update(parse_fs)
 
             # fetch ID specific metadata
             md = [{} for i in range(len(ids))]
@@ -3254,7 +3264,7 @@ html
 
         return t
 
-    def to_hdf5(self, h5grp, generated_by, compress=True):
+    def to_hdf5(self, h5grp, generated_by, compress=True, format_fs=None):
         """Store CSC and CSR in place
 
         The resulting structure of this group is below. A few basic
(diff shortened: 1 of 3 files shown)


biom/table.py
```diff
@@ -3343,6 +3353,14 @@ def to_hdf5(self, h5grp, generated_by, compress=True):
         compress : bool, optional
             Defaults to ``True`` means fields will be compressed with gzip,
             ``False`` means no compression
+        format_fs : dict, optional
+            Specify custom formatting functions for metadata fields. This dict
+            is expected to be {'metadata_field': function}, where the function
+            signature is (h5py.Group, str, dict, bool) corresponding to the
+            specific group the metadata dataset will be associated with, the
```
adamrp: "group" is slightly ambiguous in this context; maybe add "HDF5 group"?  I do recognize that the preceding signature has it listed as `h5py.Group`, but it still could be slightly confusing, IMHO.

tests/test_table.py
```diff
@@ -299,6 +299,24 @@ def test_from_hdf5_empty_md(self):
...
+        if '/' in __file__:
+            os.chdir(__file__.rsplit('/', 1)[0])
+        t = Table.from_hdf5(h5py.File('test_data/test.biom'),
+                            parse_fs=parse_fs)
+        os.chdir(cwd)
+
+        for m in t.metadata():
+            self.assertIn(m['BODY_SITE'], ('GUT', 'SKIN'))
+
+
```
adamrp: one too many blank lines between this function and the next

tests/test_table.py
```diff
@@ -299,6 +299,24 @@ def test_from_hdf5_empty_md(self):
         self.assertTrue(t._sample_metadata is None)
         self.assertTrue(t._observation_metadata is None)
 
+
```
adamrp: one too many blank lines here, too

biom/table.py
```diff
@@ -3343,6 +3353,14 @@ def to_hdf5(self, h5grp, generated_by, compress=True):
...
+            category being operated on, the metadata for the entire axis being
+            operated on, and whether to enable compression on the dataset.
+            Anything returned by this function is ignored.
```
adamrp: Might add a note that although the function does not return anything, it typically will create one or more new data sets (as is done in the tests). I was slightly confused how this was supposed to work until I looked at the test code.

tests/test_table.py
```diff
@@ -572,6 +590,42 @@ def test_to_hdf5_error(self):
...
+                             [{'barcode': 'aatt'}, {'barcode': 'ttgg'}])
+
+        def bc_formatter(grp, category, md, compression):
+            name = 'metadata/%s' % category
+            data = np.array([m[category].upper() for m in md])
+            grp.create_dataset(name, shape=data.shape, dtype=H5PY_VLEN_STR,
+                               data=data, compression=compression)
+
+
+        fname = mktemp()
```
adamrp: this function [is deprecated](https://docs.python.org/2/library/tempfile.html#tempfile.mktemp). The typical flow we have used is now:

```python
from os import close

file_handle, abs_path = mkstemp()
close(file_handle)
with open(abs_path, 'w') as new_file:
    # do stuff with new_file
```

adamrp:
:+1:

biom/table.py
wasade: done

biom/table.py
wasade: done

tests/test_table.py
wasade: done

tests/test_table.py
wasade: done

tests/test_table.py
wasade: this is the pattern in place for the rest of the hdf5 test code so would prefer to stick with it

adamrp: I still think this should be `mkstemp`, since this function (`mktemp`) was deprecated almost 12 years ago. In the entire biom-format codebase, `mktemp` is used 5 times, all of them in this file (test_table.py). I'll create an issue to replace all `mktemp`s with `mkstemp`s so that it oesn't block this PR. See #616.

wasade:
@adamrp, good to merge?

adamrp:
Yep!

PR #617, opened 2015-03-24
ENH: Replace mktemp with NamedTemporaryFile

### LINKED ISSUE #616: Replace all mktemp calls with mkstemp
The mktemp function was deprecated almost 12 years ago, and should not be used for the reasons mentioned [here](https://docs.python.org/2/library/tempfile.html#tempfile.mktemp).

The best convention to use is of the form:

```python
from os import close
from tempfile import mkstemp

file_descriptor, filepath = mkstemp()
close(file_descriptor)
with open(filepath, 'w') as new_file:
    # do stuff with new_file
```

adamrp:
Fix #616

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -11,7 +11,7 @@
 
 import os
 from json import loads
-from tempfile import mktemp
+from tempfile import NamedTemporaryFile
 from unittest import TestCase, main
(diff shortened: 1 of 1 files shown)


wasade:
thanks! will merge once tests pass

PR #618, opened 2015-03-24
MAINT: moved nested functions to module namespace

wasade:
This just shifts these methods from being defined within `Table.to_hdf5` and `Table.from_hdf5` out to the module namespace. There is no benefit in having them within the methods, and it just complicates those methods substantially. Additionally, there is a specific need within PICRUSt to be able to import the vlen formatter.


josenavas:
:+1: given tests pass

Issue #620, opened 2015-03-27
uninformative error when sorting a filtered table that is empty


jairideout:
When sorting a table that has been filtered so that it is empty (0x0), an uninformative error message is raised. This only seems to happen if the table is filtered to be empty; creating an empty `Table` outright and sorting it works.

I know this use-case sounds silly (why would you want to sort an empty table?), but this error is raised in QIIME's `observation_metadata_correlation.py` script if the user provides a metadata category that doesn't have any numeric values, resulting in a filtered table that is empty. A collaborator ran into this error and it wasn't obvious what went wrong and why.

Example:

```python
In [1]: import numpy as np

In [2]: from biom import Table

In [3]: t = Table(np.asarray([[1, 2, 3], [4, 5, 6]]), ['a', 'b'], ['c', 'd', 'e'])

In [4]: t.filter(ids_to_keep=[], axis='sample')
Out[4]: 0 x 0 <class 'biom.table.Table'> with 0 nonzero entries (0% dense)

In [5]: t.sort(sort_f = lambda _: [], axis='sample')
---------------------------------------------------------------------------
TableException                            Traceback (most recent call last)
<ipython-input-5-ebb9d7c83d49> in <module>()
----> 1 t.sort(sort_f = lambda _: [], axis='sample')

/Users/jairideout/.virtualenvs/qiime/lib/python2.7/site-packages/biom/table.pyc in sort(self, sort_f, axis)
   1753         O1  0.0 1.0 3.0
   1754         """
-> 1755         return self.sort_order(sort_f(self.ids(axis=axis)), axis=axis)
   1756
   1757     def filter(self, ids_to_keep, axis='sample', invert=False, inplace=True):

/Users/jairideout/.virtualenvs/qiime/lib/python2.7/site-packages/biom/table.pyc in sort_order(self, order, axis)
   1672                                   self.ids(axis='observation')[:], order[:],
   1673                                   self.metadata(axis='observation'), md,
-> 1674                                   self.table_id, self.type)
   1675         elif axis == 'observation':
   1676             for id_ in order:

/Users/jairideout/.virtualenvs/qiime/lib/python2.7/site-packages/biom/table.pyc in __init__(self, data, observation_ids, sample_ids, observation_metadata, sample_metadata, table_id, type, create_date, generated_by, observation_group_metadata, sample_group_metadata, **kwargs)
    258         self._observation_group_metadata = observation_group_metadata
    259
--> 260         errcheck(self)
    261
[...]

PR #621, opened 2015-03-27
620

Jorge-C:
@wasade I changed the error checking in the `filter` method so that in #620 the error would have been raised just after filtering, instead of at a later stage when sorting. Can you please check that it's fine? Thanks!

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,9 @@ Bug fixes:
 * `Table.collapse`, under `one_to_many` was not constructing the resulting
     matrix properly, issue #606
 * Improve error message when trying to load an empty file, issue #614.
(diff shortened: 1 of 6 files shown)


Jorge-C:
Hey, I've set coveralls to fail only if coverage decreases by more than 5% or below 70%. Feel free to tweak the numbers, but it's silly that we are seing build failures because a covered line was removed and that decreases coverage by 0.0001%  :)

wasade:
Agree, thanks!

Reviewing now. Please do add a changelog mention when you get a second

wasade:
This seems good, thanks! Just need the changelog mention and will merge

Jorge-C:
Thanks @wasade, changelog edited.

PR #634, opened 2015-06-03
Issue 633

### LINKED ISSUE #633: Method update_ids won't "update" ids to the same name (raises incorrect error)
When using `update_ids`, you need to provide new identifiers, otherwise the method will raise a spurious error. The example below shows how using the example table you can reproduce the error by creating a dictionary where all key and value pairs are equal:

```python
In [7]: from biom import example_table as et

In [8]: id_map = {i: i for i in et.ids()}

In [9]: id_map
Out[9]: {'S1': 'S1', 'S2': 'S2', 'S3': 'S3'}

In [10]: et.update_ids(id_map)
---------------------------------------------------------------------------
TableException                            Traceback (most recent call last)
<ipython-input-10-3d0c303209a3> in <module>()
----> 1 et.update_ids(id_map)

/Users/yoshikivazquezbaeza/.virtualenvs/qiime-191/lib/python2.7/site-packages/biom/table.pyc in update_ids(self, id_map, axis, strict, inplace)
    990                     "Mapping not provided for %s identifier: %s. If this "
    991                     "identifier should not be updated, pass strict=False."
--> 992                     % (axis, old_id))
    993             updated_ids[idx] = new_id
    994 

TableException: Mapping not provided for sample identifier: S1. If this identifier should not be updated, pass strict=False.
```

The error indicates that the mapping for `S1` was not provided, even though it is in the `id_map` dictionary. This error goes away as soon as you change the mapping so the values of the dictionary are different from the keys:

```python
In [11]: id_map = {i: i+'-:L' for i in et.ids()}

In [12]: et.update_ids(id_map)
Out[12]: 2 x 3 <class 'biom.table.Table'> with 5 nonzero entries (83% dense)
```

wasade:
Fixes #633

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   # Update conda itself
   - conda update --yes conda
 install:
-  - conda create --yes -n env_name python=$TRAVIS_PYTHON_VERSION pip numpy=$NUMPY_VERSION scipy nose pep8 Sphinx coverage
+  - conda create --yes -n env_name python=$TRAVIS_PYTHON_VERSION pip numpy=$NUMPY_VERSION scipy nose pep8 Sphinx=1.2.2 coverage
   - if [ ${USE_CYTHON} ]; then conda install --yes -n env_name cython; fi
   - if [ ${USE_H5PY} ]; then conda install --yes -n env_name h5py>=2.2.0; fi
   - source activate env_name
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,13 @@
 BIOM-Format ChangeLog
 =====================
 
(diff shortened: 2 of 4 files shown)


ElDeveloper:
LGTM! Mind adding a mention to the ChangeLog?

wasade:
After seeing [this], I'm now disappointed :)

Adding, good catch!

ElDeveloper:
Hahahaha, damn I didn't know about the world of LGTM ... thanks, I think this is ready for merge!

wasade:
@ElDeveloper, able to merge? The tests are passing

ElDeveloper:
Thanks @wasade!

wasade:
Thanks!

PR #642, opened 2015-08-28
Updated table_summarizer to remove leading space from obs

lkursell:
The legacy format for reporting ```Counts/sample detail:``` was to add a leading space in front of the sampleid, I believe for visual appeal. However, when the summary file is open in Excel, its very easy to miss that subversive space, leading to errors with my favorite VLOOKUP function. Also, this requires the user to strip the leading zero when parsing the summary file to add counts to a mapping file or other script. 

This pull request removes the leading space from the ```table_summarizer.py``` and associated ```test_table_summarizer.py```


ElDeveloper:
:+1:, this looks good to me, though I would like to have someone who was present when this initial decision was made (@wasade, @gregcaporaso or @antgonza), as I remember this was part of the now defunct `per_library_stats.py`.

The only think missing would be to add a mention to the ChangeLog.md file.

antgonza:
The code looks good, the change makes sense, and I don't see a problem by removing. However, I can't remember such conversation. Is there an open issue for this?

ElDeveloper:
@antgonza, thanks for having a look. No, there's no open issue about
this. I want to make sure that that space wasn't intended to do X or Y
that would otherwise lead to a nuclear meltdown of the microbiome forces
in life ... :grin:

wasade:
:+1:, but could the "biom_format.egg-info" bits be removed from the PR
please?

Issue #639, opened 2015-08-19
create "biom head" command to show the first n cols and rows of the table


josenavas:
Sometimes you don't know what is on the table, so seeing the first rows/cols in the summarize-table output can really useful.

antgonza:
I suggest adding another parameter not changing the current summarize-table option.

josenavas:
Make sure that the rows and columns that are shown actually include some data, if we choose the first X and Y it might not have any data, only 0.

gregcaporaso:
It might make sense to build this into the ``__repr__`` method for the class (which would then be related to https://github.com/biocore/scikit-bio/issues/656).

gregcaporaso:
This could look something like this:

```
import numpy as np

def tsv_repr(t, n):
    obs_ids = t.ids(axis='observation')[:n]
    sam_ids = t.ids(axis='sample')[:n]
    data = np.zeros((n,n))
    for i, sam_id in enumerate(sam_ids):
        for j, obs_id in enumerate(obs_ids):
            data[i,j] = t.get_value_by_ids(obs_id, sam_id)
    print('\t' + '\t'.join(sam_ids))
    for obs_id, datum in zip(obs_ids, data):
        print('\t'.join([obs_id] + map(str,list(datum))))
            
tsv_repr(t, 5)
```

<img width="759" alt="screenshot 2015-10-14 13 57 24" src="https://cloud.githubusercontent.com/assets/192372/10497593/893786ca-727b-11e5-93ac-748a5a0e64d5.png">

(but formatted nicer).

gregcaporaso:
... or maybe this would make more sense as a new ``biom head`` command?

wasade:
I like head as its not a "summarization"

gregcaporaso:
added in #669

Issue #631, opened 2015-04-24
Move CLI to click


ElDeveloper:
It would be awesome if the CLI could be moved to click, are there any current technical limitations?

wasade:
No, just dev effort and agree. I think there might already be an open issue?

ElDeveloper:
Got it, couldn't find it, I only found mentions of click in the issues.

wasade:
ah, well you take the :cake: :)

josenavas:
If we move the CLI to click means that is a major release? Just checking cause I think I won't be backwards compatible, or am I missing something?

wasade:
yes

ElDeveloper:
3.0, here we go!

wasade:
we could do a minor release i suppose

gregcaporaso:
Just remember that backward compatibility is needed for ``biom-format >= 2.1.3, < 2.2.0`` - that's what QIIME depends on.

wasade:
Ya, so I was thinking 2.2 instead of 3.0

gregcaporaso:
Ok, perfect.

ebolyen:
- [ ]   `add-metadata`        Add metadata to a BIOM table
- [ ]   `convert`             Convert to/from the BIOM table format
- [ ]   `normalize-table`     Normalize a BIOM table
- [ ]   `show-install-info`   Provide information about the biom-format installation
- [ ]   `subset-table`        Subset a BIOM table
- [ ]   `summarize-table`     Summarize sample or observation data in a BIOM table
- [ ]   `validate-table`      Validate a BIOM-formatted file

gregcaporaso:
I've tested all of the commands presented on the website, and they're working as they did in 2.1.4. Closing this.

ElDeveloper:
:fireworks:

PR #650, opened 2015-10-19
Issue639

wasade:
Partially addresses #639, but is dependent on #631 for a CLI

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -19,6 +19,7 @@ import numpy.testing as npt
 import numpy as np
 from scipy.sparse import lil_matrix, csr_matrix, csc_matrix
 
+from biom import example_table
 from biom.exception import UnknownAxisError, UnknownIDError, TableException
 from biom.util import unzip, HAVE_H5PY, H5PY_VLEN_STR
 from biom.table import (Table, prefer_self, index_list, list_nparray_to_sparse,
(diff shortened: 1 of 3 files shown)


gregcaporaso:
:+1:

tests/test_table.py
```diff
@@ -44,6 +45,49 @@
...
 class SupportTests(TestCase):
+
+    def test_head(self):
+        # example table is 2 x 3, so no change in contained data
+        exp = example_table
+        obs = example_table.head()
+        self.assertIsNot(obs, exp)
+        self.assertEqual(obs, exp)
+
+    def head_bounded(self):
```
jairideout: Should this be `test_head_bounded`?

wasade: hah, yes. good catch

josenavas:
I think I commented this in the issue but, given the inherent sparsity of the data, it will be possible that showing the first (n, m) rows/cols of data will not fulfill the goal of the command, which was to see a few non-zero values.

Probably a parameter that, if passed, searches for the first non-zero value and shows the (n, m) table starting at such value might help? e.g. if the first non-zero value is at (1, 20) it shows (1, 20) to (1+n, 20+m)

gregcaporaso:
@josenavas, I think that's a good suggestion, but let's make that change after this one goes in. This is lower priority for 2.1.5 (see the milestones for the blocking issues) so I don't want us to get bogged down in this today as we're going for the (maybe overly optimistic) goal of release today.

josenavas:
That makes sense - thanks for the clarification @gregcaporaso !!

wasade:
Agree

PR #655, opened 2015-10-19
REF: summarize-table uses click internally

ebolyen:


.travis.yml
```diff
@@ -25,6 +25,7 @@ install:
 script:
   - nosetests --with-doctest --with-coverage
   - pep8 biom setup.py
+  - flake8 biom setup.py
```
wasade: should probably run on `scripts/` as well assuming that's where the click command is going?

scripts/biomcli
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python
```
wasade: i think this needs to be named `biom` otherwise it'll break command line backwards compatibility

scripts/biomcli
```diff
@@ -0,0 +1,54 @@
...
+
+    $ biom summarize-table -i table.biom -o table_summary.txt
+
+    """
+    result = biom.cli.summarize_table(input_fp, qualitative, observations)
+    if output_fp:
+        with open(output_fp, 'w') as fh:
+            fh.write(result)
+    else:
+        click.echo(result)
```
wasade: I don't _think_ there is a test for this, but at one point we added in support so that `biom summarize-table` could be piped to other commands like `head`. It would be great if this were still supported. It probably is given click's aggressive POSIX compliance, but just making a note about it.

ebolyen:
Tests are failing because I moved things around...

.travis.yml
ebolyen: :+1:

scripts/biomcli
ebolyen: We'll rename this once we are done

scripts/biomcli
ebolyen: I can confirm locally that that still works

.travis.yml
```diff
@@ -25,6 +25,7 @@ install:
 script:
   - nosetests --with-doctest --with-coverage
   - pep8 biom setup.py
```
jairideout: remove, redundant now that we're running flake8

biom/cli/summarize_table.py
```diff
@@ -0,0 +1,99 @@
...
+
+from biom.util import biom_open, compute_counts_per_sample_stats
+from biom.parse import parse_biom_table
+from numpy import std
+from operator import itemgetter
+
+
+def summarize_table(input_fp, qualitative=False, observations=False):
+    lines = []
+    with biom_open(input_fp, 'U') as biom_f:
```
jairideout: see gitter conversation about handling this: we either implement a custom parameter type in Click to convert BIOM fp to Table, or informally make sure that each `cli` function uses `biom.load_table` to do this (for consistent handling, error msgs, etc.)

jairideout: based on gitter convo, update this to use `load_table`

scripts/biomcli
```diff
@@ -0,0 +1,54 @@
...
+
+@cli.command(name='summarize-table')
+@click.option( '-i', '--input-fp', required=True,
+              type=click.Path(exists=True, dir_okay=False),
+              help='The input BIOM table')
+@click.option('-o', '--output-fp', default=None, type=click.Path(),
+              help='An output file-path')
+@click.option('--qualitative', default=False, is_flag=True,
+              help="Present counts as number of unique observation ids per"
+                    "sample, rather than counts of observations per sample.")
```
jairideout: need a space so str join is readable

ebolyen: thanks!

.travis.yml
ebolyen: ok

biom/cli/summarize_table.py
ebolyen: ok

.travis.yml
```diff
@@ -24,7 +24,8 @@ install:
   - pip install -e . --no-deps
 script:
   - nosetests --with-doctest --with-coverage
-  - pep8 biom setup.py
+  - pep8 biom setup.py scripts
```
jairideout: remove

scripts/biomcli
```diff
@@ -0,0 +1,54 @@
...
+import biom.cli
+
+@click.group()
+@click.version_option(version=biom.__version__)
+def cli():
+    pass
+
+@cli.command(name='summarize-table')
+@click.option( '-i', '--input-fp', required=True,
+              type=click.Path(exists=True, dir_okay=False),
```
jairideout: should we add `readable=True`?

ebolyen: That's on by default

jairideout: good point sir!

PR #657, opened 2015-10-19
adds API for creating a Table from a uc file

gregcaporaso:
This code is adapted from [gregcaporaso/scaling-octo-robot](https://github.com/gregcaporaso/scaling-octo-robot), which @jairideout and I pair-programmed.

### DIFF
--- a/biom/parse.py
+++ b/biom/parse.py
@@ -246,7 +246,6 @@ def get_axis_indices(biom_str, to_keep, axis):
     return idxs, json.dumps(subset)[1:-1]  # trim off { and }
 
 
-
 def parse_uc(fh):
     """ Create a Table object from a uclust/usearch/vsearch uc file.
 
@@ -337,6 +336,7 @@ def parse_uc(fh):
             # nothing else needs to be done for 'L' records
             pass
     return Table(data, observation_ids=observation_ids, sample_ids=sample_ids)
+
 
 def parse_biom_table(fp, ids=None, axis='sample', input_is_dense=False):
     r"""Parses the biom table stored in the filepath `fp`
--- a/biom/parse.py
+++ b/biom/parse.py
@@ -15,6 +15,7 @@ from biom.exception import BiomParseException, UnknownAxisError
 from biom.table import Table
 from biom.util import biom_open, __version__
 import json
+import collections
 
 __author__ = "Justin Kuczynski"
 __copyright__ = "Copyright 2011-2013, The BIOM Format Development Team"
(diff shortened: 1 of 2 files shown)


biom/parse.py
```diff
@@ -245,6 +246,98 @@ def get_axis_indices(biom_str, to_keep, axis):
...
+
+        Raises
+        ------
+        ValueError
+            If a sequence identifier is encountered that doesn't have at least
+            one underscore in it (see Notes).
+
+        Notes
+        -----
+        This file assumes that sequence identifiers in this file are in QIIME's
```
jairideout: *This file* -> *This function*

biom/parse.py
```diff
@@ -245,6 +246,98 @@ def get_axis_indices(biom_str, to_keep, axis):
...
+
+    """
+    data = collections.defaultdict(int)
+    sample_idxs = {}
+    sample_ids = []
+    observation_idxs = {}
+    observation_ids = []
+    # The types of hit lines we need here are hit (H), seed (S) and
+    # library seed (L). Store these in a set for quick reference.
+    line_types = set(list('HSL'))
```
jairideout: don't need `list` call

tests/test_parse.py
```diff
@@ -1553,5 +1554,144 @@ def test_parse_biom_table_with_hdf5(self):
 415	0	0	0	0	0	7	0	2	2
 416	0	1	0	0	1	0	0	0	0"""
 
+
+class ParseUcTests(TestCase):
+
+        # exp = Table(np.array([[0., 1.], [3., 4.]]), ['O1', 'O2'], ['S1', 'S2'],
```
jairideout: remove

biom/parse.py
```diff
@@ -245,6 +246,98 @@ def get_axis_indices(biom_str, to_keep, axis):
...
+    for line in fh:
+        # determine if the current line is one that we need
+        line = line.strip()
+        if not (line and line[0] in line_types):
```
wasade: I think this could be expressed as a regex which would avoid the `continue`

gregcaporaso: Eh, I'd rather not adapt this for regex now. I'd prefer to just leave as it since it works.

biom/parse.py
```diff
@@ -245,6 +246,98 @@ def get_axis_indices(biom_str, to_keep, axis):
...
+        query_id = fields[8].split()[0]
+
+        if observation_id == '*':
+            # S and L lines don't have a separate observation id
+            observation_id = query_id
+
+        # get the index of the current observation id, or create it if it's
+        # the first time we're seeing this id
+        try:
+            observation_idx = observation_idxs[observation_id]
```
wasade: Since this is expected to happen frequently, could this be done as a test for membership?

biom/parse.py
```diff
@@ -245,6 +246,98 @@ def get_axis_indices(biom_str, to_keep, axis):
...
+            except ValueError:
+                raise ValueError(
+                 "A query sequence was encountered that does not have an "
+                 "underscore. An underscore is required in all query "
+                 "sequence identifiers to indicate the sample identifier.")
+            # get the sample id and its index, creating the index if it is the
+            # first time we're seeing this id
+            sample_id = query_id[:underscore_index]
+            try:
+                sample_idx = sample_idxs[sample_id]
```
wasade: same comment here w.r.t. membership

biom/parse.py
```diff
@@ -245,6 +246,98 @@ def get_axis_indices(biom_str, to_keep, axis):
...
+            except KeyError:
+                sample_idx = len(sample_ids)
+                sample_idxs[sample_id] = sample_idx
+                sample_ids.append(sample_id)
+            # increment the count of the current observation in the current
+            # sample by one.
+            data[(observation_idx, sample_idx)] += 1
+        else:
+            # nothing else needs to be done for 'L' records
```
wasade: would it be possible to have an explicit branch for this, and raise in `else`? That would add some protection against corrupt or unexpected input

biom/parse.py
```diff
@@ -245,6 +246,98 @@ def get_axis_indices(biom_str, to_keep, axis):
...
+        Notes
+        -----
+        This file assumes that sequence identifiers in this file are in QIIME's
+        "post-split-libraries" format, where the identifiers are of the form
+        ``<sample-id>_<sequence-id>``. Everything before the first underscore
+        will be used as the sample identifier in the resulting ``Table``.
+        The information after the first underscore is not used directly, though
+        the full identifiers of seeds will be used as the observation
+        identifier in the resulting ``Table``.
+
```
wasade: I thought this was going to capture stats on the hits as well too? Or is that substantially complicated that it isn't worth it?

biom/parse.py
wasade: kk

biom/parse.py
gregcaporaso: I'm not sure what stats you're wanting to collect, but no, that's out of the scope of what I'm doing with this PR.

biom/parse.py
jairideout: That would create an impossible-to-reach branch which will drop coverage.

jairideout: Note that it's guaranteed above to only be processing H, L, or S lines by this point.

biom/parse.py
wasade: The hit details that are in the uc file. This is a lossy transform at this point.

gregcaporaso: Yep (and it always has been). We can expand on this later, but I'm not going to process any of that information now.

biom/parse.py
wasade: Good point

PR #658, opened 2015-10-19
REF: shuffled API boundary for table loading

ebolyen:
Also checks that output path is writable

### DIFF
--- a/tests/test_cli/test_summarize_table.py
+++ b/tests/test_cli/test_summarize_table.py
@@ -9,6 +9,7 @@
 # -----------------------------------------------------------------------------
 
 from biom.cli import summarize_table
+from biom.parse import load_table
 
 import tempfile
 from unittest import TestCase, main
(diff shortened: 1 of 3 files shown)


tests/test_cli/test_summarize_table.py
```diff
@@ -17,14 +18,10 @@
...
     def setUp(self):
-        self.file = tempfile.NamedTemporaryFile()
-        self.file.write(biom1)
-        self.file.seek(0)
-        self.biom1 = self.file.name
-
-
-    def tearDown(self):
-        self.file.close()
+        with tempfile.NamedTemporaryFile() as fh:
```
jairideout: Probably easier to fake a file at this point, since `summarize_table` accepts a `Table` object now.

jairideout: Meh, I'm just gonna merge though, not a big deal at all

PR #659, opened 2015-10-19
REF: `biom show-install-info` uses Click

jairideout:


biom/cli/installation_informer.py
```diff
@@ -0,0 +1,99 @@
...
+    return _format_info(_get_dependency_version_info(), 'Dependency versions')
+
+
+def _get_formatted_package_info():
+    return _format_info(_get_package_info(), 'biom-format package information')
+
+
+def _get_system_info():
+    return (("Platform", sys.platform),
+            ("Python/GCC version", sys.version.replace('\n', ' ')),
```
wasade: This isn't exactly correct as I don't think we can assume all builds of Python use GCC

tests/test_cli/test_show_install_info.py
```diff
@@ -0,0 +1,24 @@
...
+# -----------------------------------------------------------------------------
+
+import unittest
+
+from biom.cli import show_install_info
+
+
+class TestShowInstallInfo(unittest.TestCase):
+    def test_default(self):
+        # Not really sure what to specifically test here, as this information
```
wasade: :+1:

biom/cli/installation_informer.py
jairideout: Note that we're just refactoring the pyqi code into click -- I'd like to avoid making changes to the code itself at this point b/c due to time constraints (there is a lot of less-than-ideal code we'll be converting over today)

wasade: Want to just ~~/GCC VERSION~~ and call it good?

jairideout: sure

jairideout: done

PR #661, opened 2015-10-19
python 3 conversion

wasade:
py3 compat with the exception of pyqi


.travis.yml
```diff
@@ -16,9 +15,10 @@ before_install:
   # Update conda itself
   - conda update --yes conda
 install:
-  - conda create --yes -n env_name python=$TRAVIS_PYTHON_VERSION pip numpy=$NUMPY_VERSION scipy nose pep8 flake8 Sphinx=1.2.2 coverage
+  - conda create --yes -n env_name python=$PYTHON_VERSION pip numpy scipy nose pep8 flake8 coverage future
   - if [ ${USE_CYTHON} ]; then conda install --yes -n env_name cython; fi
   - if [ ${USE_H5PY} ]; then conda install --yes -n env_name h5py>=2.2.0; fi
+  - if [ ${PYTHON_VERSION} = "2.7" ]; then conda install --yes -n env_name Sphinx=1.2.2; fi
   - source activate env_name
   - pip install coveralls pyqi
```
gregcaporaso: remove pyqi

gregcaporaso:
@wasade, a bunch of unit test failures for this one.

wasade:
oh yes, and they're fun

ChangeLog.md
```diff
@@ -6,13 +6,19 @@ biom 2.1.4-dev
 
 Changes:
 
+* Codebase is now Python 3 compatible.
```
ebolyen: :confetti_ball:

biom/cli/table_summarizer.py
```diff
@@ -23,7 +23,7 @@ def summarize_table(table, qualitative=False, observations=False):
         compute_counts_per_sample_stats(table, qualitative)
     num_observations = len(table.ids(axis='observation'))
 
-    counts_per_sample_values = counts_per_samp.values()
+    counts_per_sample_values = list(counts_per_samp.values())
```
ebolyen: If this is an `ndarray` `.tolist()` is much faster than using the `list` constructor.

wasade: That's a dict

biom/table.py
```diff
@@ -271,7 +274,7 @@ def vlen_list_of_str_formatter(grp, header, md, compression):
         if m[header] is None:
             continue
         value = np.asarray(m[header])
-        data[i, :len(value)] = value
+        data[i, :len(value)] = [str(v).encode('utf8') for v in value]
```
ebolyen: Just to check `v` would never be a bytestring correct?

biom/table.py
```diff
@@ -3543,7 +3554,7 @@ def axis_dump(grp, ids, md, group_md, order, compression=None):
                 # is cleaner, as is the parse
                 grp.create_dataset('ids', shape=(len_ids,),
                                    dtype=H5PY_VLEN_STR,
-                                   data=[str(i) for i in ids],
+                                   data=[str(i).encode('utf8') for i in ids],
```
ebolyen: This will do strange things (from the user's perspective) if the user hands ids as `byte`s in Python 3, but they really shouldn't do that and it works entirely as expected otherwise.

biom/table.py
wasade: It's possible. May not have back tracked this appropriately re: gutter
conversation

wasade: it would be unusual here for the same reason as you noted below on `ids`

biom/cli/table_summarizer.py
ebolyen: roger

ebolyen:
@wasade This looks good aside from `str(x)` not being a universally safe thing to do in Py3, but if we can demonstrate that it would be unreasonable to hit those code paths I think we are fine. Another way to handle would be to check for [`six.binary_type`](https://pythonhosted.org/six/#six.binary_type) and if True, then `.encode` directly, otherwise cast to `str` and encode.

PR #666, opened 2015-10-19
ENH: removed pyqi

ebolyen:
:tada: :cake: :confetti_ball:

PR #668, opened 2015-10-20
converting biom convert to be click-based

gregcaporaso:
still some items to-do - moved those to #631.


wasade:
:+1: so far

gregcaporaso:
@wasade, note one change that you may have missed: I added a new table type, ``Table``, which is the default. This makes ``--table-type`` no longer required. Since this field is never used (to the best of my knowledge) forcing users to specify it has just become annoying. This doesn't break backward compatibility. Any issues with this?

wasade:
Seems okay as the types really aren't used to the best of my knowledge

ebolyen:
flake8 error, if this gets merged I can update #666

gregcaporaso:
flake8 error should be fixed. @ebolyen, there are still some things I need to do on this one, but if you're going to work on #666 tonight you can merge this and I can take care of those items in a new PR tomorrow. I gave you write access to the repo.

ebolyen:
Thanks @gregcaporaso, I'll merge this once tests pass and then update my PR

gregcaporaso:
Thanks! I moved my to-do items to #631.

PR #669, opened 2015-10-20
Added head

wasade:
Adds the head command. This really just wraps `Table.head` so I didn't think there was a need to add in additional tests.


gregcaporaso:
Looks good, thanks!

PR #673, opened 2015-10-21
HACK: add pyqi to install_requires in Python 2

jairideout:
pyqi lives on... necessary for backward-compatibility with QIIME 1.9.x.


gregcaporaso:
So sad...

gregcaporaso:
closing in favor of #675, which has these commits and passes.

PR #684, opened 2015-11-10
BUG: fixes #683

### LINKED ISSUE #683: biom from-uc arguments
Thanks for the recent addition of the `from-uc` option. When I first tried to use it without reading the great [writeup](https://gist.github.com/gregcaporaso/f3c042e5eb806349fa18) Greg Caporaso did, I ran into an issue that I think some users *might* have too. 

When I just run `biom from-uc` it lists:
```
biom from-uc  
Usage: biom from-uc [OPTIONS]

Error: Missing option "-i" / "--input-fp"
```
If I run it with just the `-i` option I then get: `AttributeError: 'NoneType' object has no attribute 'encode'`

I recognized this is due to not having the output specified, but other users may not.

I guess this comes down to the output not being required option. Can the script do anything without specifying output, or will it always give throw `AttributeError`?

gregcaporaso:
This slipped through in the 2.1.5 release.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,9 @@ New Features:
   either axis. See [issue #645](https://github.com/biocore/biom-format/issues/639).
 * Format of numbers in ``biom summarize-table`` output is now more readable and localized. See [issue #679](https://github.com/biocore/biom-format/issues/679).
 
+Bug fixes:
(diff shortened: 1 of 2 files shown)


jairideout:
changelog?

gregcaporaso:
Added changelog mention, thanks.

jairideout:
Looks good, thanks!

Issue #690, opened 2016-01-29
biom 2.1.5 fails with umlaut


johnchase:
I was attempting to work with a biom table generated by QIIME. The table was generated using the latest [UNITE ITS](https://unite.ut.ee/repository.php) database. 

biom 2.1.4 was able to work with the table, however 2.1.5 failed

The following text is contained in the file [bad_table.txt](https://raw.githubusercontent.com/johnchase/thesis_analysis/master/bad_table.txt)
[bad_table.biom](https://github.com/johnchase/thesis_analysis/blob/master/bad_table.biom?raw=true)

```
#OTU ID sample1 taxonomy
otu1    10  s__Cortinarius grosmornënsis
```

Convert this to biom

``` bash
$biom convert -i bad_table.txt -o bad_table.biom --to-hdf5 --table-type "OTU table" --process-obs-metadata taxonomy
$biom summarize-table -i bad_table.biom

Num samples: 1
Num observations: 1
Total count: 10
Table density (fraction of non-zero values): 1
```

Upgrade to biom 2.1.5

``` bash
$pip install --upgrade biom-format
$biom summarize-table -i bad_table.biom
Traceback (most recent call last):
  File "/Users/jc33/miniconda3/envs/qiime/bin/biom", line 11, in <module>
    sys.exit(cli())
  File "/Users/jc33/miniconda3/envs/qiime/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jc33/miniconda3/envs/qiime/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/Users/jc33/miniconda3/envs/qiime/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/jc33/miniconda3/envs/qiime/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/jc33/miniconda3/envs/qiime/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/Users/jc33/miniconda3/envs/qiime/lib/python2.7/site-packages/biom/cli/table_summarizer.py", line 47, in summarize_table
    table = load_table(input_fp)
[...]

PR #691, opened 2016-01-30
utf8 files in PY2

Jorge-C:
Read biom files as if they were utf8 in PY2 (closes issue #690, but not ready for merge yet).

It just changes a few lines of code, and declares all biom text files to be encoded as utf-8 (a parameter could potentially be provided to change that).

However, it needs testing work (adding/fixing some), so please feel free to grab it and get it ready for merge in the next few days. Otherwise I'll work on it sometime next week.

Rationale for changes:
- io.open is open in PY3, and is the cleanest way to get PY3's behaviour in PY2.
- Gzipped files are trickier (they provide an encoding argument in PY3!), and it
  seems using the codecs module is the way to go [1, 2], but it needs more testing...

[1] https://docs.python.org/2/howto/unicode.html
[2] http://stackoverflow.com/a/1883734/1191119


wasade:
Hey @Jorge-C, any chance this PR can be finalized?

Jorge-C:
I'll take a look Daniel! What's your timeline?

wasade:
Nothing particular... :)

Jorge-C:
Added some tests to make sure the behaviour keeps being right in the future.

There's a test that uses `isinstance` to check that open_biom('foo.gz') returns a  `gzip.GzipFile`. To support utf-8 in py2 I need to return another duck that quacks like it... What are your thoughts about deleting that `isinstance` check?

wasade:
Am fine with dropping the `isinstance`. It looks like there are some failures btw

Jorge-C:
Yeah, the only failure was that test that used `isinstance`. This is ready now.

PR #692, opened 2016-03-03
Update `future` dependency

bgruening:
biom_format is using string_type from the future module, but this is included in version 0.15.0.

https://github.com/PythonCharmers/python-future/commit/3b9c2115aa38b516c82ff6bf370eeac31b9b6a62


ebolyen:
@bgruening Nice catch! Thanks!

PR #704, opened 2016-07-25
Rapid subset

wasade:
@josenavas, possible for a review on this? Asking as I believe you wrote the normal subset method. Basically, the subset kills you when handling metadata, which is fine for one-off runs, but I need to do repeated subsetting for EMP in the context of the block decomposition for unifrac. See [here](https://gist.github.com/wasade/60db4059e3b7e42bb648db1d10ef72d6) for the motivator, which I'd like to get running tomorrow on EMP oref. This PR doesn't block, but it's necessary for this approach to be feasible

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,13 @@ biom 2.1.5-dev
 
 Changes since biom 2.1.5 go here.
 
+New Features:
+
+* `Table.from_hdf5` now supports a rapid subset in the event that metadata is
+   not needed. In benchmarking against the Earth Microbiome Project BIOM table,
+   the reduction in runtime was multiple orders of magnitude while additionally
+   preserving substantial memory. 
+
 biom 2.1.5
 ----------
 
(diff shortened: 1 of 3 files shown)


josenavas:
Code looks good. I see in your commit history that you decided not to filter zero'd rows/cols, why was it beneficial for block decomp unifrac? I was gonna suggest adding it but if there is a good argument against it I'm fine with it :-)

wasade:
Code assumes subset is in index order with full table. Difficult to work
around right now as the skbio API takes a dense numpy matrix for counts so
no index labels

josenavas:
ooooh makes sense! Thanks @wasade

gregcaporaso:
@wasade, @josenavas is the default behavior the same? Just wondering if this will require a version bump to maintain backward compatibility for QIIME and any other projects that depend on `` < 2.2.0``. Thanks!

wasade:
Yes

gregcaporaso:
Thanks!

PR #707, opened 2016-09-22
DOC: adds PhyloToAST to software list

smdabdoub:
PhyloToAST makes extensive use of BIOM-format files in the analysis and visualization of microbiome data


wasade:
Added, thanks for the PR!

PR #715, opened 2016-10-17
Update index.rst

potter-s:
Added EBI Metagenomics


potter-s:
Added EBI Metagenomics - project uses BIOM.

wasade:
That's great, thank you!

Issue #717, opened 2016-10-26
`Table.__init__` does not properly handle `scipy.sparse.dok_matrix` 


wasade:
...or `lil_matrix`. The internals of `Table` assume either `csr` or `csc`. The constructor is expected to normalize the representation, however a `dok_matrix` evals true on the [`isspmatrix`](https://github.com/biocore/biom-format/blob/master/biom/table.py#L310) call resulting in some type confusion.

PR #720, opened 2016-11-04
Concat

### LINKED ISSUE #716: Add `Table.hstack` and `Table.vstack`
Probably `classmethod`s. Take iterables of `Table`. `vstack` asserts disjoint observations, and either subset or 100% overlap on samples. Converse for `hstack`. These operations can likely be performed more efficiently than `Table.merge` as the nz values do not need to be examined explicitly. The specific usecase is when a partitioning scheme has been performed for parallel operations, and the results need to be joined back together.

wasade:
Resolves #716. This ended up growing in complexity to handle the general case in which one set of IDs are disjoint while the other axis doesn't need to be. I'd be curious if others see ways to reduce or simplify some of the code complexity here. 

Now, why is this method useful? Often, when merging tables, one axis is disjoint (e.g., the sample IDs in a meta-analysis). This method offers 80x or so improvement over the existing `merge` method for this case. In addition, it supports concatenating multiple tables while `merge` must operate only on pairs thus offering likely further performance gains in real world scenarios.

The first benchmark shows approximately a 20x improvement. When profiling, a hotspot was identified in `sort_order`, which was also rectified.

```python
In [1]: import biom

In [2]: a = biom.load_table('twins.qiita-2014.biom')

In [3]: b = biom.load_table('ag.qiita-10317.biom')

In [4]: a
Out[4]: 8735 x 1046 <class 'biom.table.Table'> with 344079 nonzero entries (3% dense)

In [5]: b
Out[5]: 9128 x 9919 <class 'biom.table.Table'> with 1508178 nonzero entries (1% dense)

In [6]: %timeit c = a.concat([b])
1 loop, best of 3: 2.25 s per loop

In [7]: %timeit c = a.merge(b)
1 loop, best of 3: 43.9 s per loop
```

A quick profile suggested a hotspot in `sort_order` stemming from a) an implicit cast to dense (which also stood to bloat memory) and b) utilizing incremental indexing instead of fancy indexing. So I fixed that. New timing below:

```python
In [1]: import biom

In [2]: a = biom.load_table('twins.qiita-2014.biom')

In [3]: b = biom.load_table('ag.qiita-10317.biom')

In [4]: %timeit c = a.concat([b])
1 loop, best of 3: 582 ms per loop
```

### DIFF
--- a/biom/cli/table_subsetter.py
+++ b/biom/cli/table_subsetter.py
@@ -64,7 +64,10 @@ def subset_table(input_hdf5_fp, input_json_fp, axis, ids, output_fp):
             input_json_fp = f.read()
 
     with open(ids, 'U') as f:
-        ids = [line.strip() for line in f]
+        ids = []
+        for line in f:
+            if not line.startswith('#'):
+                ids.append(line.strip().split('\t')[0])
 
     table, format_ = _subset_table(input_hdf5_fp, input_json_fp, axis, ids)
 
--- a/biom/table.py
+++ b/biom/table.py
@@ -185,7 +185,7 @@ from future.utils import viewitems
 from collections import defaultdict, Hashable, Iterable
 from numpy import ndarray, asarray, zeros, newaxis
 from scipy.sparse import (coo_matrix, csc_matrix, csr_matrix, isspmatrix,
-                          vstack, hstack, lil_matrix)
+                          vstack, hstack)
 
 from future.utils import string_types
 from biom.exception import (TableException, UnknownAxisError, UnknownIDError,
--- a/biom/table.py
+++ b/biom/table.py
@@ -184,10 +184,12 @@ from future.builtins import zip
 from future.utils import viewitems
 from collections import defaultdict, Hashable, Iterable
 from numpy import ndarray, asarray, zeros, newaxis
-from scipy.sparse import coo_matrix, csc_matrix, csr_matrix, isspmatrix, vstack
+from scipy.sparse import (coo_matrix, csc_matrix, csr_matrix, isspmatrix,
+                          vstack, hstack)
 
 from future.utils import string_types
-from biom.exception import TableException, UnknownAxisError, UnknownIDError
+from biom.exception import (TableException, UnknownAxisError, UnknownIDError,
+                            DisjointIDError)
 from biom.util import (get_biom_format_version_string,
                        get_biom_format_url_string, flatten, natsort,
                        prefer_self, index_list, H5PY_VLEN_STR, HAVE_H5PY,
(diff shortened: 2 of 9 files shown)


wasade:
@antgonza relevant for qiita

@gregcaporaso relevant for `q2-feature-table` on its `merge`

wasade:
This resolves #717 as well, adding a note and test in a second.

biom/table.py
```diff
@@ -2935,6 +2921,162 @@ def _intersect_id_order(self, a, b):
...
+        # Constructed from biom file
+        #OTU ID	S1	S2	S3	S4	S5	S6	S7	S8	S9
+        O1	0.0	1.0	2.0	0.0	0.0	0.0	12.0	13.0	14.0
+        O2	3.0	4.0	5.0	0.0	0.0	0.0	0.0	0.0	0.0
+        O3	0.0	0.0	0.0	6.0	7.0	8.0	0.0	0.0	0.0
+        O4	0.0	0.0	0.0	9.0	10.0	11.0	0.0	0.0	0.0
+        O5	0.0	0.0	0.0	0.0	0.0	0.0	15.0	16.0	17.0
+
+        """
+        # should this be a staticmethod?
```
antgonza: Good question, yes? No real preference but we should remove this comment.

biom/table.py
```diff
@@ -2935,6 +2921,162 @@ def _intersect_id_order(self, a, b):
...
+                    tmp_inv_md = [None] * len(table.ids())
+                else:
+                    tmp_inv_md = list(tmp_inv_md)
+                tmp_inv_md.extend([invaxis_metadata[i] for i in missing_ids])
+
+                # resolve axis ids and metadata
+                tmp_ids = list(table.ids(axis=axis))
+                tmp_md = table.metadata(axis=axis)
+
+                # this sucks.
```
antgonza: ????????????????

wasade:
Thanks @antgonza, is this good then? I'm not thrilled with the complexity but meh...

biom/table.py
```diff
@@ -2935,6 +2921,161 @@ def _intersect_id_order(self, a, b):
...
 
+    def concat(self, others, axis='sample'):
+        """Concatenate tables if axis is disjoint
+
+        Parameters
+        ----------
+        others : iterable of biom.Table
+            Tables to concatenate
+        axis : {'sample', 'observation'}
+            The axis to concatenate on. i.e., if axis is 'sample', then tables
```
ElDeveloper: Oh, I get it now.

biom/table.py
```diff
@@ -2935,6 +2921,161 @@ def _intersect_id_order(self, a, b):
...
+
+        # verify disjoint, and fetch all ids from all tables
+        for table in all_tables:
+            table_axis_ids = table.ids(axis=axis)
+            table_invaxis_order = table.ids(axis=invaxis)
+            table_invaxis = set(table_invaxis_order)
+
+            # test we have disjoint IDs
+            if not axis_ids.isdisjoint(table_axis_ids):
+                raise DisjointIDError("IDs are not disjoint")
```
ElDeveloper: Maybe worth adding more information noting what the offending table is?

biom/table.py
```diff
@@ -2935,6 +2921,161 @@ def _intersect_id_order(self, a, b):
                 idx += 1
         return new_order
 
+    def concat(self, others, axis='sample'):
```
ElDeveloper: Not blocking, but just making an observation, this method is a little bit long, is there any way to break it down a bit?

tests/test_table.py
```diff
@@ -97,6 +99,96 @@ def test_head_zero_or_neg(self):
...
+        obs = example_table.concat([table2, ], axis='sample')
+        self.assertEqual(obs, exp)
+
+    def test_concat_raise_overlap(self):
+        with self.assertRaises(DisjointIDError):
+            example_table.concat([example_table])
+
+        with self.assertRaises(DisjointIDError):
+            example_table.concat([example_table], axis='observation')
+
```
ElDeveloper: What happens with an empty list of tables?

tests/test_table.py
```diff
@@ -1620,6 +1712,15 @@ def test_update_ids_cache_bug(self):
         exp_index = {'x': 0, 'y': 1}
         self.assertEqual(obs._sample_index, exp_index)
 
+    def test_other_spmatrix_type(self):
+        # I dont actually remember what bug stemmed from this...
```
ElDeveloper: 🤔

ElDeveloper:
This looks good, just a few minor comments. As for your original question, maybe making it a class method makes more sense. It really should just be a function somewhere, but a class method makes organizational sense 👍

wasade:
@ElDeveloper, i didn't see any comments?

I'm tempted to defer doing this as a class or staticmethod because, while not good, its more consistent for this to be a `Table` instance method...

ElDeveloper:
> I'm tempted to defer doing this as a class or staticmethod because, while not good, its more consistent for this to be a Table instance method...

Sounds good.

ElDeveloper:
Also, comments posted, that was my bad (GitHub's new UI).

biom/table.py
```diff
@@ -1795,39 +1797,23 @@ def sort_order(self, order, axis='sample'):
         O2	1.0	0.0	4.0
 
         """
-        md = []
-        vals = []
+        fancy = np.array([self.index(i, axis=axis) for i in order], dtype=int)
```
wasade: These are index positions so content of the matrix doesn't matter. Forcing `int` as numpy by default does `float`.

ElDeveloper:
Ah, ok, this makes sense!

biom/table.py
wasade: Yea, I know. Not enthusiastic. A bunch of private methods perhaps...?

biom/table.py
```diff
@@ -2935,6 +2921,161 @@ def _intersect_id_order(self, a, b):
...
         return new_order
 
+    def concat(self, others, axis='sample'):
+        """Concatenate tables if axis is disjoint
+
+        Parameters
+        ----------
+        others : iterable of biom.Table
+            Tables to concatenate
+        axis : {'sample', 'observation'}
```
wasade: done

biom/table.py
wasade: it would take a little bit of book keeping to know which two tables were in conflict, and it might be more than 2 in conflict. Blocking?

tests/test_table.py
wasade: works as expected?

```python
In [2]: biom.example_table.concat([biom.Table([], [], [])])
Out[2]: 2 x 3 <class 'biom.table.Table'> with 5 nonzero entries (83% dense)
```

ElDeveloper:
Not blocking!

ElDeveloper:
Ah, no not what I meant, I meant:

bt.concat([])

tests/test_table.py
wasade: It came up in two scenarios. I could dig through the code but not really excited about that as it would necessitate a bunch of parallel exploration in an interpreter. I could just delete the comment? :)

wasade:
Thanks!

ElDeveloper:
That sounds like a better solution haha.

ElDeveloper:
Also, looks like GitHub does not link messages together :[

wasade:
Thanks! Will test empty case and resolve remaining in about an hour

wasade:
I think that's it. Decomposition would be nice but still not really sure
what would be the best here.

antgonza:
This looks good to me. I will defer to @ElDeveloper to merge.

biom/cli/table_subsetter.py
```diff
@@ -64,7 +64,10 @@ def subset_table(input_hdf5_fp, input_json_fp, axis, ids, output_fp):
             input_json_fp = f.read()
 
     with open(ids, 'U') as f:
-        ids = [line.strip() for line in f]
+        ids = []
```
josenavas: Minor but isn't this the same?
```python
ids = [line.strip().split('\t')[0] for line in f if not line.startwith('#')
```
you know to avoid appends (no need to change if you think it makes the code less readable.

biom/table.py
```diff
@@ -2935,6 +2921,161 @@ def _intersect_id_order(self, a, b):
...
+
+        axis_ids = set()
+        invaxis_ids = set()
+        invaxis_metadata = {}
+
+        all_tables = others[:]
+        all_tables.insert(0, self)
+
+        # verify disjoint, and fetch all ids from all tables
+        for table in all_tables:
```
josenavas: I think this for loop is a good candidate for a small private function.

biom/table.py
```diff
@@ -2935,6 +2921,161 @@ def _intersect_id_order(self, a, b):
...
+        invaxis_order = sorted(invaxis_ids)
+
+        # determine what inv axis IDs do not exist per table, and pad and sort
+        # as necessary
+        padded_tables = []
+        for table in all_tables:
+            missing_ids = list(invaxis_ids - set(table.ids(axis=invaxis)))
+
+            if missing_ids:
+                # determine new shape
```
josenavas: I think the contents of this if statement can also be broken in another small private function

biom/table.py
```diff
@@ -2935,6 +2921,161 @@ def _intersect_id_order(self, a, b):
...
+                    tmp_inv_md = [None] * len(table.ids())
+                else:
+                    tmp_inv_md = list(tmp_inv_md)
+                tmp_inv_md.extend([invaxis_metadata[i] for i in missing_ids])
+
+                # resolve axis ids and metadata
+                tmp_ids = list(table.ids(axis=axis))
+                tmp_md = table.metadata(axis=axis)
+
+                # resolve construction based off axis. This really should be
```
josenavas: Would it be too much pain to put it in a classmethod?

biom/table.py
josenavas: After reading the code in here I think it makes sense to break it in a couple of private functions and probably upgrade this to a class method, it will be easier to do this change now than later and break compatibility on the API.

biom/cli/table_subsetter.py
wasade: yup, made for a long line and this isnt a performance critical block

biom/table.py
wasade: the resolution off the axis is something which definitely should be a class method and would be nice to do as it would remove this common pattern from the codebase

biom/table.py
wasade: i stepped back from wanting to make this a classmethod as, while not ideal, it would be inconsistent with the rest of the API

wasade: ...could possibly decompose...

josenavas: Consistency with the rest of the API is important so I think it is fine to leave as it is.

Just checking, do you think it is worth having an "inplace" parameter and do the modifications in place given that this is not a class method? I don't think it is necessary but wanted to ask to see what you think about it and how does it looks from an API point of view.

biom/table.py
josenavas: From your comment, it looks like this pattern is repeated in multiple parts of the code base, in such case I think it is ok to open an issue about it and fix all occurrences at once later. Do you agree?

wasade:
+1

josenavas:
@ElDeveloper wanna merge if you don't have any further comments?

Issue #731, opened 2017-02-02
biom convert to json writes the type as a byte literal


corburn:
In python 3, biom convert will write the `type` as a byte literal. As a result, the `type` is not recognize by scripts expecting a string literal. This can be reproduced with the biom-format test data where, unlike the data this was encountered in, the `id` is also written as a byte literal:
```
biom convert -i  tests/test_data/test.biom -o test.json --to-json

{
  "id": "b'No Table ID'",
  "type": "b'Unspecified'"
  ...
}
```

Including the table-type flag will replace the byte literal with a string literal.
`biom convert -i  tests/test_data/test.biom -o test.json --to-json --table-type "OTU table"`

System information
==================
         Platform:	linux
   Python version:	3.4.2 (default, Oct  8 2014, 10:45:20)  [GCC 4.9.1]
Python executable:	/biom-format/env/bin/python3

Dependency versions
===================
click version:	6.7
NumPy version:	1.12.0
SciPy version:	0.18.1
 h5py version:	2.6.0

biom-format package information
===============================
biom-format version:	2.1.5-dev

wasade:
Good catch. Am surprised this hasn't popped up or was caught by the test
harness. Will investigate today, thank you for creating the issue!

PR #732, opened 2017-02-02
Issue 731

wasade:
Resolves issue #731. Should be a quick review. We weren't enforcing a conversion from bytes to non-bytes for a few attributes.


antgonza:
what happened with the coverage?

wasade:
Might be due to the py27 failure. I hate python2.

wasade:
More travis issues? Doesn't look like the build was triggered at all

wasade:
Ya. Travis is reporting a partial system outage due to a major impact to Heroku. Lame.

wasade:
@antgonza we're green. I'm pushing in a changelog bugfix mention but other than that, I believe were good to go

Issue #708, opened 2016-09-23
biom.Table: provide API to delete axis metadata


jairideout:
`biom.Table` provides a way to add metadata (`Table.add_metadata`) and retrieve metadata (`Table.metadata`) but does not provide a convenient way to delete metadata. This would be really useful for QIIME 2. Here is a workaround provided by @wasade:

``` python
table_without_md = biom.Table(table_with_md.matrix_data,
                              table_with_md.ids(axis='observation'),
                              table_with_md.ids(axis='sample')) 
```

This workaround should be fairly efficient because the matrix data and axis IDs aren't copied; only a new `biom.Table` reference is created and some ID indexing operations are performed.

PR #737, opened 2017-03-29
More 2.1.6 milestone issues

wasade:
Closing out a few more issues for the milestone, including pandas support.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -19,6 +19,8 @@ New Features:
 * `Table.sort_order` was performing an implicit cast to dense, and not leveraging fancy indexing. A substantial performance gain was acheived. See [PR #720](https://github.com/biocore/biom-format/pull/720.
 * `biom subset-table` now accepts a QIIME-like mapping file when subsetting by IDs [Issue #587](https://github.com/biocore/biom-format/issues/587)
 * `Table.del_metadata` was added to support the removal of metadata entries from the table [Issue #708](https://github.com/biocore/biom-format/issues/708).
+* `Table.to_dataframe` was added to cast the internal matrix data to a Pandas `SparseDataFrame` [Issue #622](https://github.com/biocore/biom-format/issues/622).
+* `Table.metadata_to_dataframe` was added to cast axis metadata to a Pandas `DataFrame` [Issue #622](https://github.com/biocore/biom-format/issues/622).
 
 Bug fixes:
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -34,6 +34,7 @@ Bug fixes:
 * `Table.nnz` was not calling `eliminate_zeros()` on the underlying sparse matrix, resulting in the possibility of counting explicitly set zero values. See [#727](https://github.com/biocore/biom-format/issues/727).
 * `Table.from_hdf5` was not properly turning `bytes` into `str` for the `table_id` and the `type` HDF5 attributes. See [#731](https://github.com/biocore/biom-format/issues/731).
 * `Table.__init__` now always performs an `astype(float)` on the contained `spmatrix`. This type normalization is beneficial for underlying Cython code on the filtering and transform operations. It is possible this will introduce some performance overhead, however in _most_ cases the data should already be float. See [#718](https://github.com/biocore/biom-format/issues/718).
+* `Table.to_hdf5` was not handling lists of str appropriately in the general case. Ssee [#638](https://github.com/biocore/biom-format/issues/638).
 
 biom 2.1.5
 ----------
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -27,6 +27,7 @@ Bug fixes:
 * ``-o`` is now a required parameter of ``biom from-uc``. This was not the case previously, which resulted in a cryptic error message if ``-o`` was not provided. See [issue #683](https://github.com/biocore/biom-format/issues/683).
 * Matrices are now cast to csr on `Table` construction if the data evaluate as `isspmatrix`. This fixes [#717](https://github.com/biocore/biom-format/issues/717) where some API methods assumed the data were csc or csr.
 * `Table.concat` was not handling tables without metadata, resulting in an exception due to mismatches metadata shape. See [#724](https://github.com/biocore/biom-format/issues/724).
+* When validating a BIOM-Format 1.0.0 table, specifying the version string would trigger an error. See [#664](https://github.com/biocore/biom-format/issues/664). An explicit regression test was not added as this stemmed from an integration, and there currently is not support for script usage tests; see [#656](https://github.com/biocore/biom-format/issues/656).
 
 biom 2.1.5
 ----------
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -16,8 +16,9 @@ New Features:
   either axis. See [issue #645](https://github.com/biocore/biom-format/issues/639).
 * Format of numbers in ``biom summarize-table`` output is now more readable and localized. See [issue #679](https://github.com/biocore/biom-format/issues/679).
 * `Table.concat` has been added to the API and allows for concatenating multiple tables in which the IDs of one of the axes are known to be disjoint. This has substantial performance benefits over `Table.merge`.
-* `Table.sort_order` was performing an implicit cast to dense, and not leveraging fancy indexing. A substantial performance gain was acheived. See [PR #720](https://github.com/biocore/biom-format/pull/720)
+* `Table.sort_order` was performing an implicit cast to dense, and not leveraging fancy indexing. A substantial performance gain was acheived. See [PR #720](https://github.com/biocore/biom-format/pull/720.
 * `biom subset-table` now accepts a QIIME-like mapping file when subsetting by IDs [Issue #587](https://github.com/biocore/biom-format/issues/587)
+* `Table.del_metadata` was added to support the removal of metadata entries from the table [Issue #708](https://github.com/biocore/biom-format/issues/708).
 
 Bug fixes:
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -35,6 +35,7 @@ Bug fixes:
 * `Table.from_hdf5` was not properly turning `bytes` into `str` for the `table_id` and the `type` HDF5 attributes. See [#731](https://github.com/biocore/biom-format/issues/731).
 * `Table.__init__` now always performs an `astype(float)` on the contained `spmatrix`. This type normalization is beneficial for underlying Cython code on the filtering and transform operations. It is possible this will introduce some performance overhead, however in _most_ cases the data should already be float. See [#718](https://github.com/biocore/biom-format/issues/718).
 * `Table.to_hdf5` was not handling lists of str appropriately in the general case. Ssee [#638](https://github.com/biocore/biom-format/issues/638).
+* `Table.to_hdf5` was not handling taxonomy as flat strings, which was a common mistake that was outside of expectations for the formatter. The formatter now attempts to split on semicolon if this scenario is encountered, and errors with a more informative error if a problem occurs. See [#530](https://github.com/biocore/biom-format/issues/530).
 
 biom 2.1.5
 ----------
--- a/biom/table.py
+++ b/biom/table.py
@@ -678,25 +678,22 @@ class Table(object):
         else:
             raise UnknownAxisError("%s is not recognized" % axis)
 
-        if keys is None and axis == 'whole':
-            self._sample_metadata = None
-            self._observation_metadata = None
+        if keys is None:
+            if axis == 'whole':
+                self._sample_metadata = None
+                self._observation_metadata = None
+            elif axis == 'sample':
+                self._sample_metadata = None
+            else:
+                self._observation_metadata = None
             return
 
         for ax in axes:
             if self.metadata(axis=ax) is None:
                 continue
 
-            if keys is None:
-                keys_to_drop = set()
-                for md in self.metadata(axis=ax):
-                    keys_to_drop.update(set(md))
-                keys_to_drop = list(keys_to_drop)
-            else:
-                keys_to_drop = keys[:]
-
             for i, md in zip(self.ids(axis=ax), self.metadata(axis=ax)):
-                for k in keys_to_drop:
+                for k in keys:
                     if k in md:
                         del md[k]
 
--- a/biom/table.py
+++ b/biom/table.py
@@ -668,8 +668,8 @@ class Table(object):
         >>> tab.del_metadata(keys=['env'])
         >>> for id, md in zip(tab.ids(), tab.metadata()):
         ...     print(id, list(md.items()))
-        S1 [('barcode', 'ATGC')]
-        S2 [('barcode', 'GGTT')]
+        (S1, [('barcode', 'ATGC')])
+        (S2, [('barcode', 'GGTT')])
         """
         if axis == 'whole':
             axes = ['sample', 'observation']
--- a/biom/table.py
+++ b/biom/table.py
@@ -3899,20 +3899,18 @@ html
         expand = {}
         for key, value in kv_test:
             if isinstance(value, (tuple, list)):
-                expand[key] = len(value)
+                expand[key] = True
                 for idx in range(len(value)):
                     columns.append("%s_%d" % (key, idx))
             else:
-                expand[key] = 0
+                expand[key] = False
                 columns.append(key)
 
         rows = []
         for m in md:
             row = []
             for key, value in sorted(m.items()):
-                if expand[key] > 0:
-                    if len(value) != expand[key]:
-                        raise TypeError("'%s' contains jagged entries" % key)
+                if expand[key]:
                     for v in value:
                         row.append(v)
                 else:
--- a/biom/table.py
+++ b/biom/table.py
@@ -299,11 +299,37 @@ def vlen_list_of_str_formatter(grp, header, md, compression):
             lengths.append(len(m[header]))
 
     if not np.all(iterable_checks):
-        raise TypeError(
-            "Category %s not formatted correctly. Did you pass"
-            " --process-obs-metadata taxonomy when converting "
-            " from tsv? Please see Table.to_hdf5 docstring for"
-            " more information")
+        if header == 'taxonomy':
+            # attempt to handle the general case issue where the taxonomy
+            # was not split on semicolons and represented as a flat string
+            # instead of a list
+            def split_and_strip(i):
+                parts = i.split(';')
+                return [p.strip() for p in parts]
+            try:
+                new_md = []
+                lengths = []
+                for m in md:
+                    parts = split_and_strip(m[header])
+                    new_md.append({header: parts})
+                    lengths.append(len(parts))
+                old = deepcopy(md)  # attempt to preserve the original metadata
+                md = new_md
+            except:
+                raise TypeError("Category '%s' is not formatted properly. The "
+                                "most common issue is when 'taxonomy' is "
+                                "represented as a flat string instead of a "
+                                "list. An attempt was made to split this "
+                                "field on a ';' to coerce it into a list but "
+                                "it failed. An example entry (which is not "
+                                "assured to be the problematic entry) is "
+                                "below:\n%s" % (header, md[0][header]))
+        else:
+            raise TypeError(
+                "Category %s not formatted correctly. Did you pass"
+                " --process-obs-metadata taxonomy when converting "
+                " from tsv? Please see Table.to_hdf5 docstring for"
+                " more information" % header)
 
     max_list_len = max(lengths)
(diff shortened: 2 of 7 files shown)


wasade:
@mortonjt can you review the `DataFrame` components here?

@jairideout, can you review the `del_metadata` components here? 

@antgonza @josenavas @ElDeveloper would either of you be able to fill in the gaps?

jairideout:
@wasade is there someone else available to review `del_metadata`? I'm pretty tied up with the upcoming QIIME 2 release, so may not get to this until after the release (that'll happen in the next couple of weeks). Not sure what your release timeline is.

mortonjt:
@wasade this is awesome!  Thanks for putting this together!  I'll start testing it in a bit.

wasade:
@jairideout, this was in response to the feature request you created (#708) based on a usecase in QIIME2. I think someone else is able to review it as it isn't too complex, but as the feature was motivated by an upstream consumer, it would be ideal if someone from that team can review to assess whether its implementation is appropriate. In terms of release timing, the critical issues for this milestone are closed out by this PR so a release by the end of this week is at least possible but balanced by a current large dev push separate from this which I'm involved in. This release includes expanded functionality which QIIME2 can leverage (e.g., `Table.to_dataframe`, `Table.del_metadata`, etc) as well as a massive performance gain when merging tables (`Table.concat` for the general case of one axis which does not have overlapping IDs) so it may be something the q2 would like to consider bumping the minimum BIOM-format microversion requirement on (no API modification, just expansion).

Thanks @mortonjt!

mortonjt:
How do you want to handle the pandas dependency?  Do you want to make it required, or an optional dependency?

tests/test_table.py
```diff
@@ -1298,6 +1307,100 @@ def test_add_group_metadata_w_existing_metadata(self):
...
+        exp_samp = pd.DataFrame(['A', 'B', 'A'], index=['S1', 'S2', 'S3'],
+                                columns=['environment'])
+        exp_obs = pd.DataFrame([['Bacteria', 'Firmicutes'],
+                                ['Bacteria', 'Bacteroidetes']],
+                               index=['O1', 'O2'],
+                               columns=['taxonomy_0', 'taxonomy_1'])
+        obs_samp = example_table.metadata_to_dataframe(axis='sample')
+        obs_obs = example_table.metadata_to_dataframe(axis='observation')
+        pdt.assert_frame_equal(obs_samp, exp_samp)
+        pdt.assert_frame_equal(obs_obs, exp_obs)
```
mortonjt: Can we have an extra test case for the scenario where the taxonomies aren't complete?

For example, is it possible to return the following dataframe

```
exp_obs = pd.DataFrame([['B', 'F'], ['B'], index=['O1', 'O2'], columns=['taxonomy_0', 'taxonomy_1'])
```
where the phylum level is missing for one of the OTUs

wasade:
Required is fine

wasade:
@mortonjt, the commit I just pushed in is inaccurate, fixing right now.

mortonjt:
Awesome!  Looks like the new changes fixed the problem.  Just tested it out on a real dataset, so the metadata handling seems to be ok.

mortonjt:
I am getting some errors when I try to run `del_metadata`

```
>>> table.del_metadata(keys=['taxonomy_0'], axis='observation')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mortonjt/Dropbox/UCSD/research/biom-format/biom/table.py", line 710, in del_metadata
    i, ax, k))
KeyError: 'id TACGGAGGATGCGAGCGTTATCCGGATTTATTGGGTTTAAAGGGAGCGCAGACGGGAGATTAAGTCAGTTGTGAAAGTTTGCGGCTCAACCGTAAAATTGCAGTTGATACTGGTTTCCTTGAGTGCAGTTGAGGCAGGCGGAATTCGTGG on observation axis is missing taxonomy_0'
```
I'm assuming that should be equivalent to removing columns within the dataframe, right?

mortonjt:
Looks like the issue was the `taxaonomy_0` isn't actually in the biom file.  So the above issue isn't actually an issue.

I've tested all of the new functions, so it seems to be in working order

jairideout:
> @jairideout, this was in response to the feature request you created (#708) based on a usecase in QIIME2. I think someone else is able to review it as it isn't too complex, but as the feature was motivated by an upstream downstream consumer, it would be ideal if someone from that team can review to assess whether its implementation is appropriate.

We are working around it with [this code](https://github.com/qiime2/q2-types/blob/master/q2_types/feature_table/_transformer.py#L20-L39). To be honest, I haven't contributed to biom-format in a long time and a lot's changed in the codebase since I was an active contributor. I'll review the code but you may want someone else to take a final pass.

> This release includes expanded functionality which QIIME2 can leverage (e.g., Table.to_dataframe, Table.del_metadata, etc) as well as a massive performance gain when merging tables (Table.concat for the general case of one axis which does not have overlapping IDs) so it may be something the q2 would like to consider bumping the minimum BIOM-format microversion requirement on (no API modification, just expansion).

That's awesome! The version spec we have in the QIIME 2 packages allows the use of this new release (the spec is `>=2.1.5,<2.2.0`), so new user installs and/or upgrades will obtain this release. Just to confirm: is this release fully backwards-compatible with 2.1.5?

wasade:
@jairideout thanks! And I completely understand -- I'm mostly interested in making sure the API fits into what makes sense downstream for your usecase, which I believe it does with `Table.del_metadata(axis='whole', inplace=False)`. We can cover the nitty gritty details. I'm not aware of any 2.1.5 backwards incompatible changes. It may be worth q2 bumping from 2.1.5 to 2.1.6 when this is out though so that q2 can leverage the API expansions

biom/table.py
```diff
@@ -611,6 +639,85 @@ def add_group_metadata(self, group_md, axis='sample'):
...
+            for i, md in zip(table.ids(axis=ax), table.metadata(axis=ax)):
+                for k in keys_to_drop:
+                    if k in md:
+                        del md[k]
+                    else:
+                        raise KeyError("id %s on %s axis is missing %s" % (
+                                       i, ax, k))
+
+            # for consistency with init on absence of metadata
+            if not table.metadata(axis=ax)[0]:
```
jairideout: This is also a silent bug, related to the [issues described above](https://github.com/biocore/biom-format/pull/737/files#r108993735). This assumes that if the first metadata entry is empty, all others are empty, but this isn't guaranteed. Thus, this bug has the potential to delete more metadata than it should have. Can you also add a unit test?

tests/test_table.py
```diff
@@ -1298,6 +1307,111 @@ def test_add_group_metadata_w_existing_metadata(self):
...
+                  sample_metadata=samp_md)
+
+        exp_md = [d.copy() for d in obs_md]
+        del exp_md[0]['taxonomy']
+        del exp_md[1]['taxonomy']
+        del exp_md[2]['taxonomy']
+
+        exp = Table(d, obs_ids, samp_ids, observation_metadata=exp_md,
+                    sample_metadata=samp_md)
+        obs = t.del_metadata(keys=['taxonomy'], axis='observation')
```
jairideout: Can you add tests for `keys` that are an empty list, and a list with more than one element?

jairideout:
> @jairideout thanks! And I completely understand -- I'm mostly interested in making sure the API fits into what makes sense downstream for your usecase, which I believe it does with Table.del_metadata(axis='whole', inplace=False). We can cover the nitty gritty details.

Thanks! That API handles our use case (there are a couple of important bugs to fix though, see my review).

> I'm not aware of any 2.1.5 backwards incompatible changes. It may be worth q2 bumping from 2.1.5 to 2.1.6 when this is out though so that q2 can leverage the API expansions

Thanks for confirming! We will bump version specs to `>=2.1.6` when packages start leveraging the new APIs.

wasade:
Thank you, Jai! This is a fantastic review. I'll respond in more detail
this afternoon.

biom/table.py
```diff
@@ -611,6 +639,85 @@ def add_group_metadata(self, group_md, axis='sample'):
...
+        >>> from biom import example_table
+        >>> keys = {tuple(d) for d in example_table.metadata()]
+        >>> print(keys)
+        {('environment',)}
+        >>> example_table.add_metadata([(i, {'ex': i})
+        ...                             for i in example_table.ids()])
+        >>> keys = {tuple(sorted(d.keys())) for d in example_table.metadata()]
+        >>> print(keys)
+        {('environment', 'ex')}
+        >>> modified = example_table.del_metadata(keys=['envionment'],
```
wasade: I confess, I was falling back on travis -- I didn't have sphinx setup on my laptop and did not have internet at the time...

biom/table.py
```diff
@@ -611,6 +639,85 @@ def add_group_metadata(self, group_md, axis='sample'):
...
+            table = self
+        else:
+            table = self.copy()
+
+        for ax in axes:
+            if table.metadata(axis=ax) is None:
+                raise KeyError("%s does not contain metadata." % ax)
+
+            if keys is None:
+                test = table.metadata(axis=ax)[0]
```
wasade: With BIOM-Format 2.x, the ondisk representation forces all IDs to have common keys. However, it is accurate that the in-memory representation relies on soft integrity here, and same with BIOM-Format 1.0.0. The long lasting work around on this was to use `defaultdict`. Because of this soft integrity on the in memory object, I do agree that being conservative here is best -- good catch. And :+1: on the `keys=None` shortcut.

biom/table.py
```diff
@@ -611,6 +639,85 @@ def add_group_metadata(self, group_md, axis='sample'):
...
+                keys_to_drop = list(test.keys())
+            else:
+                keys_to_drop = keys[:]
+
+            for i, md in zip(table.ids(axis=ax), table.metadata(axis=ax)):
+                for k in keys_to_drop:
+                    if k in md:
+                        del md[k]
+                    else:
+                        raise KeyError("id %s on %s axis is missing %s" % (
```
wasade: Excellent catch. I'm just removing this exception anyway as I agree that idempotent is pleasant property here.

biom/table.py
jairideout: Ah, thanks for the correction about BIOM file format v2.x! I wasn't aware of that requirement.

biom/table.py
jairideout: Is travis running these tests though?

wasade: [yup](https://travis-ci.org/biocore/biom-format/jobs/216479298#L410)

jairideout: Thanks for confirming!

biom/table.py
```diff
@@ -611,6 +639,85 @@ def add_group_metadata(self, group_md, axis='sample'):
         else:
             raise UnknownAxisError(axis)
 
+    def del_metadata(self, keys=None, axis=None, inplace=False):
```
wasade: Done, `del_metadata` is always inplace

biom/table.py
```diff
@@ -611,6 +639,85 @@ def add_group_metadata(self, group_md, axis='sample'):
...
+            The axis to operate on. If 'whole', the operation is applied to
+            both the sample and observation axes.
+        inplace : bool, optional
+            If True, operate in place. If False, do not operate in place.
+
+        Raises
+        ------
+        KeyError
+            If a key to remove does not exist.
+            If the axis requested does not contain metadata.
```
wasade: Done

biom/table.py
wasade: Scanning over all `dict`s where appropriate now for keys

biom/table.py
wasade: Fixed

tests/test_table.py
wasade: Added

tests/test_table.py
wasade: Done

biom/table.py
```diff
@@ -611,6 +639,85 @@ def add_group_metadata(self, group_md, axis='sample'):
...
+        UnknownAxisError
+            If the requested axis does not exist.
+
+        Examples
+        --------
+        >>> from biom import example_table
+        >>> keys = {tuple(d) for d in example_table.metadata()]
```
wasade: Done

wasade:
@jairideout, I believe all of your comments have been addressed.

@antgonza @josenavas @ElDeveloper, I think we just need one high level pass now through this but I believe positioned now for merge.

biom/table.py
jairideout: > Can you add a unit test that uses all default parameters?

Can you also address this comment? Specifically, testing that `table.del_metadata()` works as expected (it used to error).

wasade: It's there, under `test_del_metadata_idempotent`

jairideout: Can you rename the test method to indicate its intention, or split it out into two methods? It's not generally good practice to test more than one thing in a unit test.

biom/table.py
```diff
@@ -611,6 +639,76 @@ def add_group_metadata(self, group_md, axis='sample'):
...
+
+            for i, md in zip(self.ids(axis=ax), self.metadata(axis=ax)):
+                for k in keys_to_drop:
+                    if k in md:
+                        del md[k]
+
+            # for consistency with init on absence of metadata
+            empties = {True if not md else False
+                       for md in self.metadata(axis=ax)}
+            if empties == {True, }:
```
wasade: Done and 100% coverage

biom/table.py
```diff
@@ -611,6 +639,76 @@ def add_group_metadata(self, group_md, axis='sample'):
...
+        ('S2', [('barcode', 'GGTT')])
+        """
+        if axis == 'whole':
+            axes = ['sample', 'observation']
+        elif axis in ('sample', 'observation'):
+            axes = [axis]
+        else:
+            raise UnknownAxisError("%s is not recognized" % axis)
+
+        if keys is None and axis == 'whole':
```
wasade: Done

wasade:
@jairideout, test was decomposed

jairideout:
Thanks @wasade! I didn't see the decomposed test that you [commented about](https://github.com/biocore/biom-format/pull/737#issuecomment-290803274), but I'll defer to you whether you want to add it or not. I've approved your changes.

wasade:
Thanks @jairideout! The decomposed tests start [here](https://github.com/biocore/biom-format/pull/737/files#diff-96c30f77a508a56e5770f99b5ab119e3R1430)

jairideout:
> Thanks @jairideout! The decomposed tests start here

I saw those new tests. The test I am specifically requesting is invoking `del_metadata` with *all* default parameters, since that invocation didn't work when I initially reviewed and tested the code. This is the exact call I'm looking for: `table.del_metadata()`. I realize that call occurs in `test_del_metadata_idempotent`, but that test is technically testing a different feature/behavior (idempotency instead of default signature), and `test_del_metadata_idempotent` could change its `del_metadata` call while still testing idempotency. That's why it's a good idea to make each test responsible for a single feature/behavior (see [ArrangeActAssert](http://wiki.c2.com/?ArrangeActAssert) pattern).

Again, up to you whether you add this or not.

wasade:
@jairideout yup, no worries, specific test is in. Thank you again for the detailed review :)

biom/table.py
```diff
@@ -3713,6 +3808,105 @@ def any_value(vals, id_, md):
 
         return t
 
+    def to_dataframe(self):
+        """Convert matrix data to a Pandas SparseDataFrame
+
+        Notes
```
ElDeveloper: Notes goes after returns, can you move this?

biom/table.py
```diff
@@ -3713,6 +3808,105 @@ def any_value(vals, id_, md):
...
+
+    def metadata_to_dataframe(self, axis):
+        """Convert axis metadata to a Pandas DataFrame
+
+        Parameters
+        ----------
+        axis : {'sample', 'observation'}
+            The axis to operate on.
+
+        Notes
```
ElDeveloper: Notes and raises goes after Returns, also, can you add a line between Raises and Returns?

biom/table.py
```diff
@@ -3713,6 +3808,105 @@ def any_value(vals, id_, md):
...
+        O2  3.0  4.0  5.0
+        """
+        # avoid dense conversion
+        # http://stackoverflow.com/a/17819427
+        mat = [pd.SparseSeries(v.toarray().ravel()) for v in self.matrix_data]
+        df = pd.SparseDataFrame(mat, index=self.ids(axis='observation'),
+                                columns=self.ids())
+        return df
+
+    def metadata_to_dataframe(self, axis):
```
ElDeveloper: Should the default axis be observation? I think that's the most common use-case.

biom/table.py
```diff
@@ -3713,6 +3808,105 @@ def any_value(vals, id_, md):
...
+        axis : {'sample', 'observation'}
+            The axis to operate on.
+
+        Notes
+        -----
+        Nested metadata (e.g., KEGG_Pathways) is not supported.
+
+        Metadata which are lists or tuples (e.g., taxonomy) are expanded such
+        that each index position is a unique column. For instance, the key
+        taxonomy will become "taxonomy_0", "taxonomy_1", etc where "taxonomy_0"
```
ElDeveloper: How terrible would it be to make these kingdom, phylum, etc? Or is there a particular reason why you've avoided that?

ChangeLog.md
```diff
@@ -16,8 +16,11 @@ New Features:
...
 * `Table.concat` has been added to the API and allows for concatenating multiple tables in which the IDs of one of the axes are known to be disjoint. This has substantial performance benefits over `Table.merge`.
-* `Table.sort_order` was performing an implicit cast to dense, and not leveraging fancy indexing. A substantial performance gain was acheived. See [PR #720](https://github.com/biocore/biom-format/pull/720)
+* `Table.sort_order` was performing an implicit cast to dense, and not leveraging fancy indexing. A substantial performance gain was acheived. See [PR #720](https://github.com/biocore/biom-format/pull/720.
 * `biom subset-table` now accepts a QIIME-like mapping file when subsetting by IDs [Issue #587](https://github.com/biocore/biom-format/issues/587)
+* `Table.del_metadata` was added to support the removal of metadata entries from the table [Issue #708](https://github.com/biocore/biom-format/issues/708).
+* `Table.to_dataframe` was added to cast the internal matrix data to a Pandas `SparseDataFrame` [Issue #622](https://github.com/biocore/biom-format/issues/622).
```
ElDeveloper: Seeing the name of these two methods, `to_dataframe` and `metadata_to_dataframe`, made me think that something more consistent would be `count_to_dataframe`, `metadata_to_dataframe`, and a new convenience method that converts a BIOM table to two dataframes for example `to_dataframes` and this returns `df_counts`, and `df_metadata`. What do you think?

ElDeveloper:
oops, my review came in late, my bad.

biom/table.py
wasade: Difficult as it makes a lot of assumptions which are not assured to be safe to make

mortonjt: I think the reason here is that there could be other metadata fields other than taxonomy.

For example, there could be KEGG families.

ChangeLog.md
wasade: `count_` isn't assured to be accurate in the case of relative frequency. I wanted to partition obtaining getting the matrix data and the metadata as often just the matrix data is desired.

biom/table.py
wasade: hmm.... okay, thanks, good catch. ill issue a quick PR on this to fix locations here and in the other places you noted. Thanks!!

ElDeveloper:
I thought that it was fairly safe if each observation was prefixed by
k__xxxxxx, p__xxxxx, etc? Is the problem validating that *all*
observations have this prefixes?

biom/table.py
ElDeveloper: But if the key of the metadata is taxonomy, then we are almost assure this is going to be the case right? It just seems like a nice amenity, but if it is not possible, then that's fine.

wasade:
Not all taxonomies follow Greengenes prefixes

ChangeLog.md
ElDeveloper: So maybe not count, but something along those lines maybe to keep the same semantics ... `matrix_to_dataframe`? It makes sense, to have the two separate (I whole heartedly agree with that) but the naming seems a bit discordant, that's why I `to_dataframes` I think makes sense.

ElDeveloper:
You are right, I was trying to convenience those of us who do use Greengenes.

ChangeLog.md
wasade: `matrix_to_dataframe` doesn't follow the semantics of `to_tsv`, `to_json` and `to_hdf5` though. The handling of metadata is also not uniform across the `to_*` methods already... :(

ElDeveloper: Ah, I see, thanks for the explanation.

jairideout:
Thanks for those updates to `del_metadata` @wasade, looks great!

PR #738, opened 2017-04-03
DOC: moving notes sections

wasade:
@ElDeveloper, able for quick review? Just moving the notes sections

### DIFF
--- a/biom/table.py
+++ b/biom/table.py
@@ -3845,6 +3845,12 @@ html
         axis : {'sample', 'observation'}
             The axis to operate on.
 
+        Returns
+        -------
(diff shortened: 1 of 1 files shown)


biom/table.py
```diff
@@ -3863,12 +3854,22 @@ def metadata_to_dataframe(self, axis):
         TypeError
             If a metadata column is a list or tuple, but is jagged over the
             axis.
+
```
ElDeveloper: Raises should go after returns, but before notes :neckbeard:

wasade:
Thanks!!

Issue #743, opened 2017-04-12
Release candidate evaluation


wasade:
- [x] build from source on OSX, verify test suite
- [x] build from source on Linux, verify test suite
- [x] on OSX and Linux, obtain some real table, enumerate the following, and verify results over both axes:
    - [x] `filter`
    - [x] `transform`, 
    - [x] `pa`
    - [x] `partition`
    - [x] `concat`
    - [x] `add_metadata`
    - [x] `del_metadata`
    - [x] `sum`
    - [x] `collapse`
    - [x] `sort_order`
    - [x] `to_hdf5`, `to_tsv`, `to_json`

- [x] on OSX and Linux, enumerate and verify the following
   - [x] `biom add-metadata`
   - [x] `biom convert` (to/from TSV, to/from JSON, to/from HDF5)
   - [x] `biom head` 
   - [x] `biom summarize-table`
   - [x] `biom normalize-table`
   - [x] `biom subset-table`
   - [x] `biom table-ids`
   - [x] `biom validate-table`
   - [x] `biom show-install-info`
   - [x] `biom from-uc`

The API exercising is now scripted out. Building a script now for enumerating the CLI.

Issue #741, opened 2017-04-12
Bump to 2.1.6


wasade:
This issue can be closed once:

- [x] code freeze
- [x] cython code regenerated
- [x] release candidate has been validated
- [x] changelog demarkation of release
- [x] version strings have been updated to 2.1.6 (`doc/conf.py` and `biom/util.py`)
- [x] copyright years have been updated if necessary
- [x] the release has been tagged
- [x] the release has been pushed into pypi
- [x] the release has been pushed into conda
- [x] gh-pages have been updated
- [x] version strings have been bumped to 2.1.6-dev
- [x] code unfreeze

PR #744, opened 2017-04-12
Release candidate testing, and release specific changes

wasade:
Provides support for #743 

This PR is also scoped now to include items on #741


wasade:
@ElDeveloper @antgonza @josenavas Possible for review? This PR catches a few items which were caught on release testing, version/copyright/etc updates, and scripts which exercise both the command line and API that were developed to facilitate release testing. It is large but I do not think this will be a complex review, and many of the LOC are due to regeneration of the Cython .c files.

Once this PR is merged, we can transition to the remaining items in #741.

biom/assets/exercise_api.py
```diff
@@ -0,0 +1,105 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2011-2015, The BIOM Format Development Team.
```
josenavas: 2015 -> 2017

Probably better to use 'current' or leave it open so we don't have to change every year?

josenavas: @wasade what is the goal of this file? I understand that it is exercising the API, but given that those calls are part of the Table object, shouldn't those functions be tested using the unit-tests of the Table object? Or am I missing something?

wasade:
Thanks!

The exercise API script is a set of integration tests. The unit tests of
course are in test_table.py

No strong feelings on year or current

wasade:
C files are auto generated

PR #750, opened 2017-05-22
Added documentation to the table class in table.py

### LINKED ISSUE #749: Table.__init__ does not have a docstring
Would be good if it included a description of the parameters, etc, like with other methods.

@nreeve17 will take care of this.

nreeve17:
closes #749

### DIFF
--- a/biom/table.py
+++ b/biom/table.py
@@ -353,6 +353,76 @@ class Table(object):
 
     Give in to the power of the Table!
 
+    Creates a table object in support of the BIOM file format. BIOM version
+    1.0 is based on JSON to provide the overall structure for the format
+    while versions 2.0 and 2.1 are based on HDF5. For more information see [1]_
+    and [2]_
+
+    Paramaters
+    ----------
+    data : array_like,
+        An (N,M)sample by observation matrix represented as one of these
+        types:
+            An 1-dimensional array of values
+            An n-dimensional array of values
+            An empty list
+            A list of numpy arrays
+            A list of dictionaries
+            A list of sparse matrices
+            A dictionary of values
+            A list of lists
+            A sparse matrix of values
+    observation_ids : array_like of str
+        A (N,) dataset of the observation IDs, where N is the total number
+        of IDs
+    sample_ids : array_like of str
+        A (M,) dataset of the sample IDs, where M is the total number of IDs
+    observation_metadata : list of dicts
+        per observation dictionary of annotations where every key represents a
+        metadata field that contains specific metadata information,
+        ie taxonomy, KEGG pathway, etc
+    sample_metadata : array_like of dicts
+        per sample dictionary of annotations where every key represents a
+        metadata field that contains sample specific metadata information, ie
+    table_id : str or None, Optional
+        A field that can be used to identify the table
+    type : {None, "OTU table", "Pathway table", "Function table",
+            "Ortholog table", "Gene table", "Metabolite table", "Taxon table"}
+        The type of table represented
+    create_date : str, optional
+        Date that this table was built
+    generated_by : str, optional
+        Individual who built the table
+    observation_group_metadata : list
+        group that contains observation specific group metadata information
+        (e.g., phylogenetic tree)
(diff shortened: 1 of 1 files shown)


biom/table.py
```diff
@@ -353,6 +353,76 @@ class Table(object):
 
     Give in to the power of the Table!
 
+    Creates a table object in support of the BIOM file format. BIOM version
```
wasade: "Creates an in-memory representation of a BIOM file."

biom/table.py
```diff
@@ -353,6 +353,76 @@ class Table(object):
...
+    1.0 is based on JSON to provide the overall structure for the format
+    while versions 2.0 and 2.1 are based on HDF5. For more information see [1]_
+    and [2]_
+
+    Paramaters
+    ----------
+    data : array_like,
+        An (N,M)sample by observation matrix represented as one of these
```
wasade: space between ) and sample

biom/table.py
```diff
@@ -353,6 +353,76 @@ class Table(object):
...
+        types:
+            An 1-dimensional array of values
+            An n-dimensional array of values
+            An empty list
+            A list of numpy arrays
+            A list of dictionaries
```
wasade: dictionaries -> dict

biom/table.py
```diff
@@ -353,6 +353,76 @@ class Table(object):
...
+            A list of sparse matrices
+            A dictionary of values
+            A list of lists
+            A sparse matrix of values
+    observation_ids : array_like of str
+        A (N,) dataset of the observation IDs, where N is the total number
+        of IDs
+    sample_ids : array_like of str
+        A (M,) dataset of the sample IDs, where M is the total number of IDs
+    observation_metadata : list of dicts
```
wasade: can this be noted as optional? eg `observation_metadata : list of dicts, optional`

wasade: ...same for the other optional fields

biom/table.py
```diff
@@ -353,6 +353,76 @@ class Table(object):
...
+        per observation dictionary of annotations where every key represents a
+        metadata field that contains specific metadata information,
+        ie taxonomy, KEGG pathway, etc
+    sample_metadata : array_like of dicts
+        per sample dictionary of annotations where every key represents a
+        metadata field that contains sample specific metadata information, ie
+    table_id : str or None, Optional
```
wasade: Optional -> optional

biom/table.py
```diff
@@ -353,6 +353,76 @@ class Table(object):
...
+        group that contains observation specific group metadata information
+        (e.g., phylogenetic tree)
+    sample_group_metadata : list
+        group that contains sample specific group metadata information
+        (e.g., relationships between samples)
+
+    Attributes
+    ----------
+    data
+    observation_ids
```
wasade: these are not attributes, but are accessed via `Table.ids()`

biom/table.py
```diff
@@ -353,6 +353,76 @@ class Table(object):
...
+    sample_ids
+    observation_metadata
```
wasade: same

biom/table.py
```diff
@@ -353,6 +353,76 @@ class Table(object):
...
+    sample_metadata
+    table_id
+    type
+    observation_group_metadata
```
wasade: same i think... im not aware of anyone that actually uses the group metadata

nreeve17:
this should be ready for review thanks!

wasade:
Thanks @nreeve17. I apologize for the delayed review. There is one build error (see [here](https://travis-ci.org/biocore/biom-format/jobs/237711668#L320)). I'm :+1: after that is resolved

nreeve17:
The build error, is that the same error that is causing things to crash
that use Python 3.5 and 3.6?

wasade:
It's a flake8 error so it will trigger all builds to fail -- I haven't verified that is the only error across all the builds however but it probably is.

wasade:
Thanks @nreeve17!

PR #757, opened 2017-09-27
BUG: All ints should be accepted

ElDeveloper:
This had the consequence of flagging tables as invalid when nnz values were
stored as np.int32 or np.int16 (somehow this happened for a Qiita table).

The format document specifies that this value should be int, but it doesn't specify
the specific representation, so I assume all subdtypes of np.integer should be safe.

### DIFF
--- a/biom/cli/table_validator.py
+++ b/biom/cli/table_validator.py
@@ -343,8 +343,8 @@ class TableValidator(object):
             return key
 
     def _is_int(self, x):
-        """Return True if x is an int"""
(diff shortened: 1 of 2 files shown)


wasade:
Thanks! The table was valid otherwise...? Was this table produced by qiita?

ElDeveloper:
Yeah, the table was valid otherwise. Yes, it was (study 1629 if you are curious-study is public)

ElDeveloper:
Thanks!

PR #769, opened 2018-01-22
Some additional error checking on to_hdf5

wasade:
Some additional error checking when writing out to HDF5, with some additional care on the general formatter. 

The metadata are still quite awkward, and this further underscores the need to transition the metadata representation to `pandas`.


stevendbrown:
Is there anything I can do to help get this over the line?

wasade:
Thank you for the offer! I think it will be easiest right now for me to just finish off this PR which hopefully isn't much more prodding

Issue #774, opened 2018-04-20
Add sampling with replacement to Table.subsample


wasade:
This could likely adapt the strategy used in [scikit-bio](https://github.com/biocore/scikit-bio/blob/107fc8fba3df69ceea6dd3ef357212eecc1f9be3/skbio/stats/_subsample.py#L241) and is important for instances where the sum of a vector is massive. 

Cc @mortonjt (couldn't assign for some reason)

mortonjt:
Quick question on reproducibility

Looking at the tests for [`Table.subsample`]https://github.com/biocore/biom-format/blob/master/tests/test_table.py#L2619)  I don't see any uses of numpy random seeds.  And I'm having trouble making consistent unittests when using `Table.subsample`.

@wasade , any thoughts on setting random seeds?

wasade:
No strong feelings

stevendbrown:
Do either of you have a pre-PR branch for this that I could try to pick up and carry forward?

wasade:
@stevendbrown, thank you for the inquiry! If you have bandwidth, we'd love a PR. The change should be relatively small, as it should just require a branch and call to `np.random.multinomial` as done [here](https://github.com/biocore/scikit-bio/blob/107fc8fba3df69ceea6dd3ef357212eecc1f9be3/skbio/stats/_subsample.py#L240-L241). I don't think this needs to be done within the Cython code. This would also be a pleasant feature expansion for [`q2-feature-table`](https://github.com/qiime2/q2-feature-table/blob/master/q2_feature_table/_subsample.py) and I would commit to making sure your contribution is available in the next release of QIIME 2.

stevendbrown:
@wasade OK, I made a basic implementation but I did it in the Cython code since that logic is already worked out, and doesn't require using dense data. I'm checking my work now to make sure I didn't bungle something non-obviously, but on the surface it looks pretty clean. Are there reasons I shouldn't do this and should do it instead using pure Python (e.g. maybe writing tests is harder)?

mortonjt:
Note that the numpy implementation of multinomial is already written in C -- not entirely sure how much faster it will be in a cython implementation ...

stevendbrown:
[...]

PR #783, opened 2018-09-04
addition of subsampling with replacement

stevendbrown:
This PR addresses issue #774 by adding subsampling with replacement. Following convention from scikit-bio, the sum of counts in a sample need not be <= `n` if `with_replacement == True`. I haven't added tests for this yet, but I was going to take the lead for that again from scikit-bio.


biom/_subsample.pyx
```diff
@@ -47,17 +47,21 @@ def _subsample(arr, n):
...
-        r = np.arange(length, dtype=np.int32)
-        unpacked = np.repeat(r, data_i[start:end])
-        permuted = np.random.permutation(unpacked)[:n]
         
-        result = np.zeros(length, dtype=np.float64)
-        for idx in range(permuted.shape[0]):
-            result[permuted[idx]] += 1
+        if not with_replacement:
+            if counts_sum < n:
+                data[start:end] = 0
```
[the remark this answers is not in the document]
wasade: No, this is w/o replacement.

biom/table.py
```diff
@@ -2697,6 +2697,10 @@ def subsample(self, n, axis='sample', by_id=False):
             matrix (e.g., rarefaction). If `True`, the subsampling is based on
             the IDs (e.g., fetch a random subset of samples). Default is
             `False`.
+        with_replacement : boolean, optional
+            If `False` (default), subsample without replacement. If `True`,
+            resample with replacement via the multinomial distribution.
+            Ignored when `by_id` is `True`.
```
wasade: Can this situation raise an exception instead?

biom/_subsample.pyx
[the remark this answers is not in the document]
stevendbrown: Yes, that should be what happens. The conditional is `if not with_replacement` which would be `False` if `with_replacement == True`. I agree that this is confusing; I will change the ordering of the if/else block to have `if with_replacement` at the start and the without-replacement code in the `else` block.

wasade:
The test failures in CI look like sphinx or flake8 version changes and appear unrelated to the PR. We'll need to issue a fix in a separate PR.

stevendbrown:
Hopefully this resolves earlier comments. Also added my name to `__credits__`.

biom/table.py
```diff
@@ -2700,7 +2700,7 @@ def subsample(self, n, axis='sample', by_id=False, with_replacement=False):
         with_replacement : boolean, optional
             If `False` (default), subsample without replacement. If `True`,
             resample with replacement via the multinomial distribution.
-            Ignored when `by_id` is `True`.
+            If `True` and `by_id=True`, a ValueError is raised.
```
wasade: Sorry to be a stickler, but can you use the numpydoc structure? See `table.max` for an example of describing the exceptions

biom/table.py
```diff
@@ -2697,6 +2697,10 @@ def subsample(self, n, axis='sample', by_id=False):
             matrix (e.g., rarefaction). If `True`, the subsampling is based on
             the IDs (e.g., fetch a random subset of samples). Default is
             `False`.
+        with_replacement : boolean, optional
+            If `False` (default), subsample without replacement. If `True`,
+            resample with replacement via the multinomial distribution.
+            If `True` and `by_id=True`, a ValueError is raised.
```
wasade: Had a comment here but it seemed to drop -- apologize if this is a duplicate. Can you please add the exception note to the `Raises` section instead?

tests/test_table.py
```diff
@@ -2696,6 +2696,58 @@ def test_subsample_md_copy_bug(self):
...
+        possible subsamplings.
+        """
+        a = np.array([2, 1, 2, 1, 8, 6, 3, 3, 5, 5]).T
+        dt = Table(data=a, sample_ids=['S1',],
+                   observation_ids=['OTU{:02d}'.format(i) for i in range(10)])
+        actual = set()
+        for i in range(1000):
+            obs = dt.subsample(35)
+            actual.add(tuple(obs.data('S1')))
+        self.assertTrue(len(actual) == 10)
```
wasade: This should be `self.assertEqual(len(actual), 10)`

tests/test_table.py
```diff
@@ -2696,6 +2696,58 @@ def test_subsample_md_copy_bug(self):
...
+        possible subsamplings.
+        """
+        a = np.array([2, 1, 2, 1, 8, 6, 3, 3, 5, 5]).T
+        dt = Table(data=a, sample_ids=['S1',],
+                   observation_ids=['OTU{:02d}'.format(i) for i in range(10)])
+        actual = set()
+        for i in range(1000):
+            obs = dt.subsample(35, with_replacement=True)
+            actual.add(tuple(obs.data('S1')))
+        self.assertTrue(len(actual) > 10)
```
wasade: This should be `self.assertGreaterThan(len(actual), 10)`

tests/test_table.py
```diff
@@ -2696,6 +2696,58 @@ def test_subsample_md_copy_bug(self):
...
+
+    def test_subsample_with_replacement_n(self):
+        dt = Table(np.array([[0,1,2], [2,0,1], [1, 2, 0]]),
+                   ['O1', 'O2', 'O3'],
+                   ['S1', 'S2', 'S3'])
+        new_dt = dt.subsample(20, with_replacement=True)
+        new_counts = np.unique(new_dt.sum('sample'))
+        self.assertEqual(new_counts.shape[0], 1)
```
wasade: This would always be the case though, right? The `sum` reduces the matrix to a vector

stevendbrown: I would certainly hope, yes, but it won't be if the number of unique values in the `sum` result vector is greater than 1, which again it should not be. If the subsampling ever returns different values for different samples (e.g. [20., 20., 21.]) then `new_counts.shape[0]` would be `2`.

Or do you mean `len(new_counts.shape)` should always be `1` because it gets reduced to a vector?

biom/table.py
stevendbrown: I think this should be the case now.

tests/test_table.py
wasade: My bad, I thought numpy was expressing the shape as a two element tuple, but you are correct! Sorry about that

wasade:
Excellent. I think this is essentially good. Last request, would it be possible to add an entry into the changelog noting this new feature and with a hyperlink to the issue describing it?

I'm working with @ElDeveloper right now on getting the CI issues cleaned up as they're similar to issues we've seen over time on a few other projects. Once those are resolved, I'll restart the tests here.

Thanks again @stevendbrown!

tests/test_table.py
stevendbrown: Running `nosetests` for me throws an error here:

`AttributeError: 'SparseTableTests' object has no attribute 'assertGreaterThan'`

I'm looking into where a declaration needs to be made, but if you know the answer off the top of your head, I'd be grateful to know it too.

stevendbrown:
`ChangeLog.md` has been updated, and the only real test error remaining (for me) is a failure to find `assertGreaterThan`. Working on it.

tests/test_table.py
stevendbrown: Nevermind, I found it. It's `assertGreater` rather than `assertGreaterThan`.

stevendbrown:
`nosetests` passes for me without errors.

wasade:
@stevendbrown, can you pull master? We just fixed the flake8 / doctest issues

stevendbrown:
@wasade Hopefully I've accomplished my rebasing goal here.

wasade:
In the future, just FYI, it should be feasible to just:

```bash
$ git checkout your_branch
$ git pull upstream master
```

Assuming `upstream` is the repo you forked from. The changes now associated w/ this PR are a bit broader than I think you were expecting, but I'm okay with that for this repo.

wasade:
Thank you for the contribution!!!

stevendbrown:
Thanks! I was surprised by how GitHub handled it, but when I look at the most recent commit to `biocore:master`, the actual code changes attributed to this PR are only the ones related to subsampling. Clearly I am still a `git` newbie. So much to learn, so little time.

PR #784, opened 2018-09-05
replace bare excepts with non-bare or more specific checks

stevendbrown:
Since a lot of errors on current PRs seem to stem from E722 ('do not use bare except'), I'm trying to rectify those situations by replacing the try-except blocks with more specific checks, or at least adding `BaseException`.

### DIFF
--- a/biom/cli/table_validator.py
+++ b/biom/cli/table_validator.py
@@ -401,7 +401,7 @@ class TableValidator(object):
                 datetime.strptime(val, fmt)
                 valid_time = True
                 break
-            except:
+            except BaseException:
                 pass
 
         if valid_time:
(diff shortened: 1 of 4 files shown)


stevendbrown:
@wasade Is Python 2.x still actively supported by this project? This PR passes all Travis-CI checks for Python 3.x; it's going to take me a while longer to fix doctest for the 2.x->3.x NumPy output gap.

wasade:
@stevendbrown, I apologize for not letting you know, we're in the process of resolving these outstanding issues in #769. Please do not worry about the doctest numpy output gap.

I'm in principle okay with killing off py2 support however that would trigger a minor version as opposed to micro version release.

PR #786, opened 2018-09-10
Issue762

wasade:
@johnchase, this was originally opened by you, able to review by chance? Should be quick


tests/test_table.py
```diff
@@ -1398,6 +1398,13 @@ def test_to_dataframe(self):
         obs = example_table.to_dataframe()
         pdt.assert_frame_equal(obs, exp)
 
+    def test_to_dataframe(self):
```
ElDeveloper: Can you rename this method?

johnchase:
@wasade This all looks good to me. This will be a nice addition

wasade:
@ElDeveloper, possible to merge?

ChangeLog.md
```diff
@@ -6,10 +6,13 @@ biom 2.1.6-dev
 
 Changes since 2.1.6 go here.
 
+Important: Python 3.4 support has been dropped. We now only support Python 2.7, 3.5 and 3.6.
+
```
ElDeveloper: Would also be good to add a note about the pandas versions that are supported.

setup.py
```diff
@@ -89,7 +89,7 @@
 # In brief, numpy 1.14.0 introduced whitespace changes that impact repr
 # on doctest
 install_requires = ["click", "numpy >= 1.9.2, < 1.14.0", "future >= 0.16.0",
-                    "scipy >= 0.13.0", 'pandas >= 0.19.2',
+                    "scipy >= 0.13.0", 'pandas >= 0.20.0',
```
ElDeveloper: Just noticed that .travis.yml is not going to honor this version restriction since conda is installing pandas and biom is installed with the `--no-deps` flag. I think the Travis file needs to change.

Issue #808, opened 2019-03-06
[python] `to_dataframe` does not produce sparse data frames


cdiener:
Hi,

I noticed that the `pandas.SparseDataFrame` returned by `Table.to_dataframe` is not really sparse. For instance for the American Gut data:

```python3
In [15]: bm = load_table("deblur_125nt_no_blooms.biom")

In [16]: bm
Out[16]: 32954 x 9511 <class 'biom.table.Table'> with 1829490 nonzero entries (0% dense)

In [17]: tab = bm.to_dataframe()

In [19]: type(tab)
Out[19]: pandas.core.sparse.frame.SparseDataFrame

In [20]: tab.density
Out[20]: 1.0

In [21]: tab.info()
<class 'pandas.core.sparse.frame.SparseDataFrame'>
Index: 32954 entries, AACGTAGGGTGCAAGCGTTATCCGGATTTACTGGGTGTAAAGGGAGCGCAGGCGGAAGGCTAAGTCTGATGTGAAAGCCCGGGGCTCAACCCCGGTACTGCATTGGAAACTGGTCATCTAGAGTG to TACGGGGGATGCGAGCGTTATCCGGATTCATTGGGTTTAAAGGGTGCGCAGGCCGAGGTTCAAGTCAGCGGTGAAACCCCCGCGCTCAACGCGGGGCATGCCGTTGATACTGTATCTCTGGAGTA
Columns: 9511 entries, 10317.000012326 to 10317.000038478
dtypes: Sparse[float64, nan](9511)
memory usage: 2.3+ GB
``` 

This is basically the memory use of the full table including zeros. Also the densities of the original table and the `SparseDataTable` are pretty different (~0% vs 100%).

wasade:
Interesting. So, unlike `scipy.sparse`, pandas expects empty values to be `nan` and not zero. That's super annoying.

Do you have a fix by chance?

```python
In [1]: import biom

In [2]: print(biom.example_table)
# Constructed from biom file
#OTU ID	S1	S2	S3
O1	0.0	1.0	2.0
O2	3.0	4.0	5.0

In [3]: biom.example_table.to_dataframe()
Out[3]:
     S1   S2   S3
O1  0.0  1.0  2.0
O2  3.0  4.0  5.0

In [4]: biom.example_table.to_dataframe().info()
<class 'pandas.core.sparse.frame.SparseDataFrame'>
Index: 2 entries, O1 to O2
Data columns (total 3 columns):
[...]

PR #809, opened 2019-03-07
make `Table.to_dataframe` create real sparse frames

cdiener:
So this is a proposed fix to #808. It basically specifies the fill value for sparse data explicitly. I also added a test to check for that in the future. 

Before that the the matrix data was first converted to individual `pandas.SparseSeries` and passed to `pandas.SparseDataFrame`. This is now done directly without the intermediate allocations since `SparseDataFrame` accepts any scipy sparse matrix. If `Table.matrix_data` can be something else than a numpy matrix or scipy sparse matrix this might not work but for now all tests seem to pass. Happy to change that back to the old behavior if that was on purpose.

As a side effect `to_dataframe` is now much faster on large sparse data sets. Takes about 20s on my machine for the American Gut biom (was more than 30m before).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -10,21 +10,23 @@ New Features:
 
 Bug fixes:
 
+* `Table.to_dataframe(dense=False)` does now correctly produce sparse data frames (and not accidentally dense ones as before)
+
 biom 2.1.7
 ----------
 
 New features and bug fixes, released on 28 September 2018.
 
-Important: 
+Important:
 
 * Python 3.4 support has been dropped. We now only support Python 2.7, 3.5, 3.6 and 3.7.
 * We will be dropping Python 2.7 support on the next release.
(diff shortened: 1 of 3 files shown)


wasade:
This is great, thank you! In prior versions of pandas, we couldn't pass
scipy sparse objects. Do you know by chance if this effects our present
minimum version requirement?

cdiener:
[Was introduced in pandas 0.20.0](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.20.0.html?highlight=sparsedataframe#scipy-sparse-matrix-from-to-sparsedataframe), so should be fine with the current version requirement (>=0.20.0).

ElDeveloper:
Looks great!

wasade:
Thanks @cdiener and @ElDeveloper!

PR #817, opened 2019-06-06
feat: allow for direct tsv output with optional progress bar

wwydmanski:
Handling big files is very uncomfortable in current version - converting .biom to tsv requires reading all of the data into memory and only then writing it to file. Now it is possible to do so on-the-fly, with a neat TQDM progress bar (specified by `verbose` parameter)

### DIFF
--- a/biom/table.py
+++ b/biom/table.py
@@ -187,6 +187,7 @@ from numpy import ndarray, asarray, zeros, newaxis
 from scipy.sparse import (coo_matrix, csc_matrix, csr_matrix, isspmatrix,
                           vstack, hstack)
 import pandas as pd
+import tqdm
 
 import six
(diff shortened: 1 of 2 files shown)


wwydmanski:
Checks are failing because of tqdm dependency, I'm not sure what I should do about it

wasade:
Thanks, @wwydmanski. Would it be possible to use the [progressbar](https://click.palletsprojects.com/en/5.x/api/#click.progressbar) from `click`? The major advantage is that we already depend on click. I am concerned though with mixing UI logic into backend. Would it be possible to modify the code to yield elements out to the existing `click` [interface](https://github.com/biocore/biom-format/blob/master/biom/cli/table_converter.py#L94)?

PR #829, opened 2020-01-04
BLD: test support for py38

wasade:
Testing whether travis builds support py38


wasade:
This is a minor change to the build matrix, so just merging w/o review.

PR #833, opened 2020-01-13
Docs to py37

wasade:
Docs fail on py36 but seem to work on py37 (locally). Ultimately, we should update from sphinx 1.2.2 to something much more modern, but that is another issue for another day.


wasade:
Commit history is weird, closing and reissuing

PR #836, opened 2020-01-18
feat: allow for direct tsv output

wwydmanski:
Handling big files is very uncomfortable in current version - converting .biom to tsv requires reading all of the data into memory and only then writing it to file. Now it is possible to do so on-the-fly


biom/table.py
```diff
@@ -1599,20 +1599,32 @@ def to_utf8(i):
...
             obs_id = to_utf8(obs_id)
             if header_key and obs_metadata is not None:
                 md = obs_metadata[self._obs_index[obs_id]]
                 md_out = metadata_formatter(md.get(header_key, None))
-                output.append(
-                    u'%s%s%s\t%s' %
-                    (obs_id, delim, str_obs_vals, md_out))
+                if direct_io is None:
+                    output.append(
+                        u'%s%s%s\t%s' %
```
wasade: Would it be possible to do this formatting and construction a single time prior to the if statement evaluation?

biom/table.py
```diff
@@ -1599,20 +1599,32 @@ def to_utf8(i):
...
+                        (obs_id, delim, str_obs_vals, md_out))
+                else:
+                    direct_io.write(u'%s%s%s\t%s\n' %
+                                    (obs_id, delim, str_obs_vals, md_out))
             else:
-                output.append(u'%s%s%s' % (obs_id, delim, str_obs_vals))
+                if direct_io is None:
+                    output.append(u'%s%s%s' % (obs_id, delim, str_obs_vals))
```
wasade: Same here

biom/table.py
```diff
@@ -4815,6 +4829,10 @@ def to_tsv(self, header_key=None, header_value=None,
         observation_column_name : str, optional
             Defaults to "#OTU ID". The name of the first column in the output
             table, corresponding to the observation IDs.
+        direct_io : file or file-like object, optional
+            Defaults to ``None``. Must implementing a ``write`` function. If
```
wasade: Implementing -> implement

wasade:
Thanks for the contribution @wwydmanski!

wwydmanski:
> Would it be possible to add an entry into the ChangeLog as well?
Should I add it to the section with "New Features" of 2.1.8 version?

wasade:
Good question. I think the 2.1.8 section would be best -- I'm in the middle of fixing the release, which had an unfortunate deployment bug, and I think it would be fine to include this.

Issue #837, opened 2020-02-08
Test issue with pandas1.0


tillea:
Hi,
the Debian packaged biom-format received a [bug report](https://bugs.debian.org/950929) in connection with pandas 1.0:

    ======================================================================
    ERROR: test_to_dataframe (biom.tests.test_table.TableTests)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File
    "/build/python-biom-format-2.1.8+dfsg/.pybuild/cpython3_3.8_biom-format/build/biom/tests/test_table.py",
    line 1476, in test_to_dataframe
        exp = pd.SparseDataFrame(np.array([[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]),
    TypeError: SparseDataFrame() takes no arguments

    ======================================================================
    ERROR: test_to_dataframe_is_sparse (biom.tests.test_table.TableTests)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File
    "/build/python-biom-format-2.1.8+dfsg/.pybuild/cpython3_3.8_biom-format/build/biom/tests/test_table.py",
    line 1484, in test_to_dataframe_is_sparse
        df = example_table.to_dataframe()
      File
    "/build/python-biom-format-2.1.8+dfsg/.pybuild/cpython3_3.8_biom-format/build/biom/table.py",
    line 4053, in to_dataframe
        return constructor(mat, index=index, columns=columns)
    TypeError: SparseDataFrame() takes no arguments

    ----------------------------------------------------------------------
    Ran 343 tests in 5.493s

Could you please have a look into this?
Kind regards, Andreas.

mortonjt:
@wasade fyi - I'm also able to reproduce this issue in the mmvec tutorials.

mr-c:
Hello. This bug has resulted in biom-format being [removed from Debian's testing distribution](https://tracker.debian.org/news/1172668/python-biom-format-removed-from-testing/), and it is likely to result in complete removal from Debian if not fixed. Any progress on this?

Many thanks,

wasade:
Hi @mr-c, this has been fixed (see this [comment](https://github.com/biocore/biom-format/pull/838#issuecomment-687295879, and commits cf1901c and 87dbdec). I didn't see the cross reference to this issue when working on the other PR, and am closing it now. If I cut a release today, could it be restored in Debian's testing distribution?

mr-c:
[...]

PR #838, opened 2020-02-23
Make to_dataframe use the new Sparse data structures from pandas >= 0.25

fedarko:
Currently, pandas v1 is the reason folks are running into problems like #837. This is also breaking a couple of other packages that rely on biom, including but probably not limited to [songbird](https://github.com/biocore/songbird/issues/117) and [mmvec](https://github.com/biocore/mmvec/issues/123).

This PR should fix the problem for _most_ datasets. However, **there are a few complicating factors that I think will prevent merging this PR in right now**, both due to [`pd.DataFrame.sparse.from_spmatrix()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sparse.from_spmatrix.html) (which is pandas' recommended way to convert SciPy CSR matrices to DataFrames-with-sparse-values):

1. ~~I'm pretty sure `from_spmatrix()` doesn't handle the sparse data in these matrices properly. When I make a huge but super sparse matrix in SciPy and try to convert it to a DF, my computer explodes at the conversion step (even though initializing the matrix takes on the order of 15-ish seconds for SciPy). I've opened https://github.com/pandas-dev/pandas/issues/32196 accordingly.~~ **Update:** this assumption was incorrect on my part, and it's expected that really wide matrices like this really will slow down pandas. See the pandas issue thread linked for profiling details / their response.

   To ensure that this change doesn't cause a repeat of #808, I added a test in this PR that makes sure that these sorts of large matrices can be converted to sparse DataFrames in a sane amount of time (max 30 seconds). This currently fails, since `biom.table.Table.to_dataframe()` in this PR uses `from_spmatrix()` under the hood. Making `to_dataframe()` faster should cause this test to pass. (If I'm being overzealous and sparse DataFrames are really going to be slower to create than I'd thought, it's totally possible to bump up the 30-second timer in this test.) **Update:** will need to do this and/or just try a less crazy sparse matrix than a 1mil x 1mil one.

2. ...Also I think `from_spmatrix()` has [other bugs](https://github.com/pandas-dev/pandas/issues/28992) that might make it unreliable.

ANYWAY TLDR hopefully these changes are useful at least as a starting point for a solution to this problem. Thanks!

### DIFF
--- a/biom/table.py
+++ b/biom/table.py
@@ -178,7 +178,7 @@ import scipy.stats
 from copy import deepcopy
 from datetime import datetime
 from json import dumps
-from functools import reduce, partial
+from functools import reduce
 from operator import itemgetter
 from future.builtins import zip
 from future.utils import viewitems
@@ -4027,16 +4027,19 @@ html
         return t
 
     def to_dataframe(self, dense=False):
-        """Convert matrix data to a Pandas SparseDataFrame or DataFrame
+        """Convert matrix data to a Pandas DataFrame
 
         Parameters
         ----------
         dense : bool, optional
-            If True, return pd.DataFrame instead of pd.SparseDataFrame.
+            If True, return a DataFrame that doesn't use sparse arrays
+            internally.
+            The default behavior (if dense is False) is to return a DataFrame
+            that uses sparse arrays.
 
         Returns
         -------
-        pd.DataFrame or pd.SparseDataFrame
+        pd.DataFrame
             A DataFrame indexed on the observation IDs, with the column
             names as the sample IDs.
 
@@ -4061,9 +4064,7 @@ html
             constructor = pd.DataFrame
         else:
             mat = self.matrix_data
-            constructor = partial(pd.SparseDataFrame,
-                                  default_fill_value=0,
-                                  copy=True)
+            constructor = pd.DataFrame.sparse.from_spmatrix
 
         return constructor(mat, index=index, columns=columns)
 
--- a/biom/tests/test_table.py
+++ b/biom/tests/test_table.py
@@ -1497,12 +1497,22 @@ class TableTests(TestCase):
         pdt.assert_frame_equal(obs, exp)
 
     def test_to_dataframe_is_sparse_and_efficient(self):
-        # Make a 1 million x 1 million table, where the only entries are on the
-        # diagonal. Stored dense, this would have 1 trillion entries, so it'd
-        # explode most computers (or at least take a lot of memory to use);
-        # stored sparse, this should only require enough memory to keep track
-        # of the 1 million entries along the diagonal.
+        """Tests that to_dataframe() is efficient for large/sparse tables."""
+
         def _make_big_table_and_try_conversion():
+            """Creates a very sparse/large table and calls to_dataframe on it.
+
+               If things go well, this shouldn't take more than a few seconds
+               -- but if a dense representation of the table is being used
+               internally somewhere, this will likely cause memory problems
+               that will trigger a timeout of this test.
+            """
+            # We're going to make a 1 million x 1 million table, where the only
+            # entries are on the diagonal. Stored dense, this would have
+            # 1 trillion entries, so it'd explode most computers (or at least
+            # take a lot of memory to use); stored sparse, this should only
+            # require enough memory to keep track of the 1 million entries
+            # along the diagonal.
             mil = 1000000
             big_csr_diag_1s = csr_matrix((mil, mil), dtype="float")
             big_csr_diag_1s.setdiag(1)
@@ -1510,9 +1520,9 @@ class TableTests(TestCase):
             sample_ids = ["S{}".format(i) for i in range(mil)]
             tbl = Table(big_csr_diag_1s, feature_ids, sample_ids)
             tbl.to_dataframe()
-            raise ValueError
-        # Process code based on example in
-        # https://medium.com/@chaoren/how-to-timeout-in-python-726002bf2291
+
+        # Process code based on example at
+        # https://medium.com/@chaoren/how-to-timeout-in-python-726002bf2291.
         p = Process(target=_make_big_table_and_try_conversion)
         p.start()
         # Block until 30 seconds have passed.
--- a/biom/tests/test_table.py
+++ b/biom/tests/test_table.py
@@ -1500,8 +1500,8 @@ class TableTests(TestCase):
         # Make a 1 million x 1 million table, where the only entries are on the
         # diagonal. Stored dense, this would have 1 trillion entries, so it'd
         # explode most computers (or at least take a lot of memory to use);
-        # stored dense, this should only require enough memory to keep track of
-        # the 1 million entries along the diagonal.
+        # stored sparse, this should only require enough memory to keep track
+        # of the 1 million entries along the diagonal.
         def _make_big_table_and_try_conversion():
             mil = 1000000
             big_csr_diag_1s = csr_matrix((mil, mil), dtype="float")
--- a/biom/tests/test_table.py
+++ b/biom/tests/test_table.py
@@ -11,6 +11,7 @@ from json import loads
 from tempfile import NamedTemporaryFile
 from unittest import TestCase, main
 from io import StringIO
+from multiprocessing import Process
 
 import six
 import numpy.testing as npt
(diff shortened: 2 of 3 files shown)


wasade:
Thanks, @fedarko and thank you for the detailed exploration here. The observation with `from_spmatrix` in (2) is very concerning. I believe we need to wait on upstream to address that issue.

ebolyen:
Re: 2, is this not an issue with pandas < 1.0? 
I was planning to do an skbio release to get pandas 1.0 in for QIIME 2, but if `from_spmatrix` is broken, then we might need to just pin to the older pandas for a bit longer.

ebolyen:
Ah, from the linked PR in songbird, that appears to be the case.

fedarko:
@ebolyen I haven't checked if (2) is a problem with pandas < 1.0 -- the reason songbird (and a few other things) were breaking IIRC is that biom.Table.to_dataframe() calls pandas.SparseDataFrame() directly, which was removed in pandas 1.0.

ebolyen:
I will try to reproduce in the last non-1.0 version to see what's going on.

ebolyen:
Yep, this is an issue in our last release also (2019.10):
```pytb
In [1]: from scipy.sparse import csr_matrix                                     

In [2]: import pandas as pd                                                     


In [3]: a = csr_matrix([[1, 0, 0], [2, 2, 0], [0, 2, 2]])                       

In [4]: a[0, 2] = 0                                                             


In [5]: a.todense()                                                             
Out[5]: 
matrix([[1, 0, 0],
        [2, 2, 0],
        [0, 2, 2]], dtype=int64)

In [6]: pd.DataFrame.sparse.from_spmatrix(a)                                    
Out[6]: 
   0  1  2
0  1  0  0
1  2  2  0
2  0  2  0

In [7]: pd.__version__                                                          
Out[7]: '0.25.2'

```

---

To what extent does any of our code mutate the CSR matrix before loading it as a df? @wasade do we need a workaround for this?

wasade:
@ebolyen, the `.matrix_data` member is mutable...

ebolyen:
It doesn't look like any q2 code attempts mutation, the most common path is `toarray()` which shouldn't have this issue.

<details>
<summary>grep qiime2</summary>

```
§ grep -r '.matrix_data' *
paper2/notebooks/qiime2-protocol-API.ipynb:      "  for r in self.matrix_data.tocsr()]\n",
q2-diversity/build/lib/q2_diversity/_alpha/_method.py:    counts = table.matrix_data.toarray().astype(int).T
q2-diversity/build/lib/q2_diversity/_alpha/_method.py:    counts = table.matrix_data.toarray().astype(int).T
q2-diversity/build/lib/q2_diversity/_beta/_method.py:    counts = table.matrix_data.toarray().astype(int).T
q2-diversity/build/lib/q2_diversity/_beta/_method.py:    counts = table.matrix_data.toarray().astype(int).T
q2-diversity/q2_diversity/_alpha/_method.py:    counts = table.matrix_data.toarray().astype(int).T
q2-diversity/q2_diversity/_alpha/_method.py:    counts = table.matrix_data.toarray().astype(int).T
q2-diversity/q2_diversity/_beta/_method.py:    counts = table.matrix_data.toarray().T
q2-diversity-lib/q2_diversity_lib/alpha.py:    counts = presence_absence_table.matrix_data.toarray().astype(int).T
q2-diversity-lib/q2_diversity_lib/alpha.py:    counts = presence_absence_table.matrix_data.toarray().astype(int).T
q2-diversity-lib/q2_diversity_lib/alpha.py:    counts = table.matrix_data.toarray().T
q2-diversity-lib/q2_diversity_lib/alpha.py:    counts = table.matrix_data.toarray().T
q2-feature-table/q2_feature_table/tests/test_transform.py:        npt.assert_array_equal(a.matrix_data.toarray(),
q2-feature-table/q2_feature_table/tests/test_transform.py:        npt.assert_array_equal(a.matrix_data.toarray(),
q2-gamma/q2_gamma/visualizers/plot/visualizer.py:    array = table.pa().matrix_data.astype(np.uint8).toarray()
q2-sample-classifier/q2_sample_classifier/utilities.py:    for i, row in enumerate(feature_data.matrix_data.T):
q2-types/q2_types/feature_data/tests/test_transformer.py:        table = biom.Table(table.matrix_data,
q2-types/q2_types/feature_data/tests/test_transformer.py:        table = biom.Table(table.matrix_data,
q2-types/q2_types/feature_table/_transformer.py:    return biom.Table(table.matrix_data,
q2-types/q2_types/feature_table/_transformer.py:    array = table.matrix_data.toarray().T
```
</details>

<details>
<summary>grep skbio</summary>

```
§ grep -rI '.matrix_data' *
biom-format/biom/table.py:    def matrix_data(self):
biom-format/tests/test_table.py:        obs = self.simple_derived.matrix_data
biom-format/tests/test_table.py:            self.simple_derived.matrix_data = 'foo'
biom-format/ChangeLog.md:* Matrix data can now be accessed by the ``Table.matrix_data`` property
biom-format/ChangeLog.md:* added ``Table.matrix_data`` property
scikit-bio/build/lib.linux-x86_64-3.7/skbio/stats/ordination/_principal_coordinate_analysis.py:    matrix_data = center_distance_matrix(distance_matrix.data, inplace=inplace)
scikit-bio/build/lib.linux-x86_64-3.7/skbio/stats/ordination/_principal_coordinate_analysis.py:        if method == "fsvd" and matrix_data.shape[0] > 10:
scikit-bio/build/lib.linux-x86_64-3.7/skbio/stats/ordination/_principal_coordinate_analysis.py:        number_of_dimensions = matrix_data.shape[0]
scikit-bio/build/lib.linux-x86_64-3.7/skbio/stats/ordination/_principal_coordinate_analysis.py:        eigvals, eigvecs = eigh(matrix_data)
scikit-bio/build/lib.linux-x86_64-3.7/skbio/stats/ordination/_principal_coordinate_analysis.py:        eigvals, eigvecs = _fsvd(matrix_data, number_of_dimensions)
scikit-bio/build/lib.linux-x86_64-3.7/skbio/stats/ordination/_principal_coordinate_analysis.py:        sum_eigenvalues = np.trace(matrix_data)
scikit-bio/skbio/stats/ordination/_principal_coordinate_analysis.py:    matrix_data = center_distance_matrix(distance_matrix.data, inplace=inplace)
scikit-bio/skbio/stats/ordination/_principal_coordinate_analysis.py:        if method == "fsvd" and matrix_data.shape[0] > 10:
scikit-bio/skbio/stats/ordination/_principal_coordinate_analysis.py:        number_of_dimensions = matrix_data.shape[0]
scikit-bio/skbio/stats/ordination/_principal_coordinate_analysis.py:        eigvals, eigvecs = eigh(matrix_data)
scikit-bio/skbio/stats/ordination/_principal_coordinate_analysis.py:        eigvals, eigvecs = _fsvd(matrix_data, number_of_dimensions)
scikit-bio/skbio/stats/ordination/_principal_coordinate_analysis.py:        sum_eigenvalues = np.trace(matrix_data)
unifrac/unifrac/tests/test_api.py:        cnts = table_inmem.matrix_data.astype(int).toarray().T

```
</summary>
</details>

---
It's a bummer, but I don't think we've been directly impacted by this (yet at least).

wasade:
`.transform()`, for exampel, can operate [in place](https://github.com/biocore/biom-format/blob/master/biom/_transform.pyx#L51) on the CSR/CSC representation. I'm unsure about impact to QIIME 2 -- hopefully most operations rely on the `Table` object rather than the `DataFrame`. I cannot predict the impact for other users of `biom`.

mortonjt:
One thing to note is that many users who are using the dataframe conversion
rely on dense dataframes rather than sparse dataframes.
In Songbird, we do the conversion from sparse to dense versions anyways.
One possibility is to open up the option to do an explicit dense conversion
without the an intermediate sparse dataframe (with the warning that the
memory consumption maybe large).

wasade:
Dense representations were explicitly removed from BIOM a long time ago. I'm not very eager to add support back in because a dependency has a bug.

wasade:
@fedarko, I ended up just putting in the `from_spmatrix` piece similar to what's done here within #848. It looks like this is going to be or was fixed in `Pandas`...

PR #845, opened 2020-08-11
add to_anndata conversion

### LINKED ISSUE #844: Feature request: AnnData converesion
Hello,

I'm wondering if there is any interest in adding a feature for a conversion tool from biom data to the [AnnData](https://anndata.readthedocs.io/en/latest/) format. This format is used by many tools for the analysis of single-cell RNA-sequencing data, which to my understanding, can be generally applied to microbiome data (e.g., our tool for dimensionality reduction that @mortonjt [made a tutorial for](https://github.com/mortonjt/scVI/blob/biom/tests/notebooks/biom_tutorial.ipynb)).

It seems like AnnData has the same general structure as biom so I don't believe it should be a difficult addition.

adamgayoso:
Adds `to_anndata` method to `Table` with optional anndata dependency.

- Fixes #844


.travis.yml
```diff
@@ -14,9 +14,10 @@ install:
   - source activate env_name
   - if [ ${PYTHON_VERSION} = "3.7" ]; then pip install sphinx==1.2.2 "docutils<0.14"; fi
   - pip install coveralls
+  - pip install anndata
```
wasade, line 17: Would it be possible to set this up as an optional build for the build matrix? (e.g., in `env` put `- PYTHON_VERSION=3.6 WITH_ANNDATA=True`, and then use a conditional here for install?

biom/table.py
```diff
@@ -4067,6 +4067,52 @@ def to_dataframe(self, dense=False):
...
+        >>> from biom import example_table
+        >>> adata = example_table.to_anndata()
+        >>> adata
+        AnnData object with n_obs × n_vars = 3 × 2
+            obs: 'environment'
+            var: 'taxonomy_0', 'taxonomy_1'
+        """
+        try:
+            import anndata
+        except ImportError:
```
wasade, line 4102: :+1:

biom/tests/test_table.py
```diff
@@ -1493,6 +1493,27 @@ def test_to_dataframe_dense(self):
         obs = example_table.to_dataframe(dense=True)
         pdt.assert_frame_equal(obs, exp)
 
+    def test_to_anndata_dense(self):
+        exp = pd.DataFrame(np.array([[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]),
```
wasade, line 1497: I _think_ this is equivalent to `exp = example_table.to_dataframe(dense=True)`

biom/tests/test_table.py
```diff
@@ -1493,6 +1493,27 @@ def test_to_dataframe_dense(self):
...
+                           index=['O1', 'O2'],
+                           columns=['S1', 'S2', 'S3'])
+        adata = example_table.to_anndata(dense=True)
+        pdt.assert_frame_equal(adata.transpose().to_df(), exp.astype(np.float32))
```
wasade, line 1501: anndata is `float32`?

setup.py
```diff
@@ -138,7 +138,8 @@ def run_tests(self):
       ext_modules=extensions,
       include_dirs=[np.get_include()],
       install_requires=install_requires,
-      extras_require={'hdf5': ["h5py >= 2.2.0"]
+      extras_require={'hdf5': ["h5py >= 2.2.0"],
```
wasade, line 141: ah, this is vestigial... h5py is required. But, that does not need to be addressed in this PR>

biom/tests/test_table.py
adamgayoso: It is by default. I just added an option and will update this test.

.travis.yml
adamgayoso: wouldn't this make it fail for the other versions?

biom/tests/test_table.py
adamgayoso: done

biom/tests/test_table.py
wasade: Okay cool :) no worries just curious

.travis.yml
wasade: Ah, good point. Since this is an optional dependency, it would be good to setup the tests with a skip decorator -- sorry for not catching that on the initial review. See [here](https://github.com/biocore/biom-format/blob/master/biom/tests/test_table.py#L584) as an example of how it could be done.

adamgayoso: please let me know if you like the new solution.

wasade:
@adamgayoso the failure is unrelated. I'm exploring this in #846.

adamgayoso:
Thanks @wasade. Do you need anything else from me?

wasade:
Hi @adamgayoso, I don't believe so.

wasade:
@adamgayoso, sorry for the delay. The build failures with pandas are now fixed, but it looks like there may be some flake8 [issues](https://travis-ci.com/github/biocore/biom-format/jobs/370789603#L535-L537) with this PR. Able to fix by chance?

adamgayoso:
@wasade Done. Not so happy with what I had to do with the unused import in the test, not sure if you know another way.

wasade:
Thank you!! I think that's fine, an alternative would be to specify `# noqa`, but effect is the same here.

PR #847, opened 2020-08-13
PERF: Optimize remove_empty

ElDeveloper:
This uses vectorized operations to figure out which rows/columns to
remove. Takes processing a 12,000 x 200,000 table from 20 seconds to 1.1
seconds.


biom/table.py
```diff
@@ -3216,7 +3216,7 @@ def remove_empty(self, axis='whole', inplace=True):
             axes = [axis]
 
         for ax in axes:
-            table.filter(lambda v, i, md: (v > 0).sum(), axis=ax)
+            table.filter(table.ids(axis=ax)[table.sum(axis=ax) > 0], axis=ax)
```
wasade, line 3219: I _think_ this should be cast to a `set`, e.g: table.filter(set(table.ids(axis=ax)[table.sum(axis=ax) > 0]), axis=ax)`

wasade: I might have misremembered here as it looks like a general iterable just works :)

ElDeveloper: 👍

wasade:
So am +1 but able to add a changelog mention? 

The failure looks unrelated. Am working on addressing in #846.

wasade:
@ElDeveloper, able to resolve the conflict by chance?

ElDeveloper:
Should be resolved, let's wait for the CI build to pass.

wasade:
Thanks!!

PR #848, opened 2020-09-03
Fast merge

wasade:
Added a (much) faster way to merge tables when the axes overlap. `Table.concat` is still (much) faster for the special case where one axis is assured to be disjoint. This revised merge method can operate on many tables at once, unlike the existing `merge` which took a single table. 

I don't have exact timing information, but the existing merge with four tables, each with roughly 600 samples and 1M features, was around 3 hours. This new method took a few minutes.

### DIFF
--- a/biom/table.py
+++ b/biom/table.py
@@ -3584,7 +3584,11 @@ class Table(object):
         """
         s_md = self.metadata()
         o_md = self.metadata(axis='observation')
-        if s_md is None and o_md is None:
+        no_md = (s_md is None) and (o_md is None)
+        ignore_md = (sample_metadata_f == None) and \
+                (observation_metadata_f == None)
+
+        if no_md or ignore_md:
             if sample == 'union' and observation == 'union':
                 if isinstance(other, (list, set, tuple)):
                     return self._fast_merge(other)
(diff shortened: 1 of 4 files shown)


biom/table.py
```diff
@@ -3457,6 +3457,67 @@ def concat(self, others, axis='sample'):
...
+        # aggregate the values where the outer and inner values in the
+        # multiindex are the same
+        collapsed_rcv = grouped.groupby(level=[0, 1]).sum()
+
+        # convert into a representation understood by the Table constructor
+        list_list = [[r, c, v] for (r, c), v in collapsed_rcv.items()]
+
+        return self.__class__(list_list, feature_order, sample_order)
+
     def merge(self, other, sample='union', observation='union',
```
antgonza, line 3521: Will it be worth adding a flag to `merge` to ignore metadata; this will be helpful if we wanted to force using the fast merge and our tables have metadata and we want to ignore.

antgonza (commented):
@wasade, looks great, thank you! One question.

biom/table.py
wasade: Oh, ya, that's a really good idea. Will do

antgonza:
@wasade, thank you! Looks good!

PR #850, opened 2020-11-03
Adjacency to biom

### LINKED ISSUE #823: Import into biom from standard adjacency two column table
Useful for inputs from metabolomics data. 

I can address this issue, spoke to Daniel earlier this week about the issue.

wasade:
Fixes #823 

Provides support for parsing adjacency data into `biom.Table`. The key functional addition here is that these data often are described in terms of identifier / identifier / value (e.g., "`featureX\tsampleY\t3.0`"), and this provides support to map those data directly into a sparse matrix and subsequent representation as `biom.Table`.


biom/table.py
```diff
@@ -4737,6 +4737,95 @@ def to_json(self, generated_by, direct_io=None):
...
+            assert len(parts) == 3
+            observations.append(parts[0])
+            samples.append(parts[1])
+            values.append(float(parts[2]))
+
+        # determine a stable order and index positioning for the identifiers
+        obs_order = sorted(set(observations))
+        samp_order = sorted(set(samples))
+        obs_index = {o: i for i, o in enumerate(obs_order)}
+        samp_index = {o: i for i, o in enumerate(samp_order)}
```
ElDeveloper, line 4819:
```suggestion
        samp_index = {s: i for i, s in enumerate(samp_order)}
```

biom/table.py
```diff
@@ -4737,6 +4737,95 @@ def to_json(self, generated_by, direct_io=None):
...
+        >>> func = lambda x : x
+        >>> test_table = Table.from_adjacency(data_fh)
+        """
+        if not isinstance(lines, (list, tuple)):
+            if hasattr(lines, 'readlines'):
+                lines = lines.readlines()
+            elif hasattr(lines, 'splitlines'):
+                lines = lines.splitlines()
+            else:
+                raise ValueError("Not sure how to handle this input")
```
ElDeveloper, line 4778: Can you add a test for this exception?

biom/table.py
```diff
@@ -4737,6 +4737,95 @@ def to_json(self, generated_by, direct_io=None):
...
+
+        Examples
+        --------
+        Parse tab separated adjacency data into a table:
+
+        >>> from biom.table import Table
+        >>> from io import StringIO
+        >>> data = 'a\\tb\\t1\\na\\tc\\t2\\nd\\tc\\t3'
+        >>> data_fh = StringIO(data)
+        >>> func = lambda x : x
```
ElDeveloper, line 4769:
```suggestion
```

biom/table.py
```diff
@@ -4737,6 +4737,95 @@ def to_json(self, generated_by, direct_io=None):
...
+            start, stop = match.span()
+            if (stop - start) == len(item):
+                return True
+            else:
+                return False
+
+        # sanity check and determine if we have a header or not
+        lh = lines[0].strip().split('\t')
+        if len(lh) != 3:
+            raise ValueError("Does not appear to be an adjacency format")
```
ElDeveloper, line 4793: Can you add a test for this  exception?

biom/table.py
```diff
@@ -4737,6 +4737,95 @@ def to_json(self, generated_by, direct_io=None):
...
+        elif lh == ['#OTU ID', 'SampleID', 'value']:
+            include_line_zero = False
+        elif not is_num(lh[0]) and not is_num(lh[1]) and is_num(lh[2]):
```
ElDeveloper, line 4796: From what I gather here only values are allowed to be numbers? Aren't there cases where OTU ids or sample ids could look like numbers but really be identifiers instead? I think this would trigger an error for a valid line, right?

ElDeveloper (changes requested):
Just a few comments.

biom/table.py
wasade: Sounds good, will do

biom/table.py
wasade: Sounds good, will do

biom/table.py
wasade: Fair. So maybe allow cols 1 and 2 to be anything, and only assert column 3 is numeric?

biom/table.py
```diff
@@ -4737,6 +4737,96 @@ def to_json(self, generated_by, direct_io=None):
...
+        -------
+        biom.Table
+            A BIOM ``Table`` object
+
+        Notes
+        -----
+        The input is expected to be of the form: observation, sample, value. A
+        header is not required, but if present, it must be of the form:
+
+        #OTU ID<tab>SampleID<tab>value
```
ElDeveloper, line 4759:
```suggestion
        #OTU ID<tab>SampleID<tab>value
        
        Raises
        ------
        ValueError
            If the input is not an iterable or file-like object.
        ValueError
            If the data is incorrectly formatted.
```

ElDeveloper (approved):
Looks good just one minor suggestion otherwise ready to merge.

wasade:
woo! thanks @ElDeveloper!

PR #851, opened 2020-11-03
Simplify concat

wasade:
This PR allows for (a) calling `biom.Table.concat` with a single table as well as an iterable of tables and (b) provides a helper method (`biom.concat`) that removes the need for the common pattern `table = tables[0].concat(tables[1:])` in favor of `table = biom.concat(tables)`


biom/__init__.py
```diff
@@ -73,5 +73,10 @@
                        {'environment': 'A'}], input_is_dense=True)
 
 
+def concat(tables, *args, **kwargs):
+    """Wrapper for biom.Table.concat which requires a table instance"""
```
ElDeveloper, line 77: Can you add the little bit of numpy doc here? Seems like a very handy function that would be good to have well documented.

biom/__init__.py
```diff
@@ -73,5 +73,10 @@
                        {'environment': 'A'}], input_is_dense=True)
 
 
+def concat(tables, *args, **kwargs):
```
ElDeveloper, line 76: Can you add a test for this function?

ElDeveloper (changes requested):
A few small comments, looks good and helpful!!

biom/__init__.py
wasade: yup, can do!

biom/__init__.py
wasade: Ya sounds good

ElDeveloper (approved):

wasade:
Thanks!!

PR #853, opened 2020-11-16
Normalize metadata on parse

wasade:
During QIIME 2 2020.11 environment testing, @thermokarst observed an unusual test failure in qiime2/q2-deblur where a parsed BIOM-Format v2.1.0 table contained ASCII strings rather than UTF-8 strings. This appears to be due to a possible variation in the underlying hdf5 or h5py dependency. The PR here is defensive, where on parse it casts bytes -> UTF8. Note that variable length lists (e.g., taxonomy) already decode to UTF-8.


thermokarst (approved):
Tested locally (in situ on busywork, more specifically), can confirm that this addresses the failures in https://busywork.qiime2.org/teams/main/pipelines/master-distribution/jobs/unit-test-q2-deblur/builds/118, on linux-64 and darwin.

PR #865, opened 2021-11-18
Add support for Python 3.10

nsoranzo:
Also:
- Cleanup pre-3.6 code with pyupgrade
- Add requirement for `>=3.6` to `setup.cfg`


wasade:
Thank you!!! This is very helpful! I'm :+1: assuming CI is green, I just approved the deploy

nsoranzo:
Thanks @wasade ! For reference, I've added the minimal patch to support Python 3.10 to https://github.com/conda-forge/biom-format-feedstock/pull/17 .

wasade:
I saw :) and thank you for that as well :)

wasade:
@nsoranzo any chance you could modify gh actions to include 3.7-3.10? (https://github.com/biocore/biom-format/blob/master/.github/workflows/python-package-conda.yml#L49) Just saw it was pinned, don't recall why, and i'm curious what happens here. If failures do arise, I'll still merge this PR as it's green, and I'll open a separate gh issue to resolve them.

nsoranzo:
I normally find that just min and max supported versions are sufficient (i.e. ['3.6', '3.10']), but up to you!

nsoranzo:
The 3.10 job is failing I think because there is no `scikit-bio` conda package for Python 3.10

wasade:
Okay, thanks. That would explain it, and I know the team is working on a new release. I don't think that is a real dependency, or at least I'm not finding imports for it anymore -- want to remove it from the gh actions conda install? (i'd suggest/commit but gh isn't letting me do so in the pr)

nsoranzo:
It was added in https://github.com/biocore/biom-format/pull/859 and it's still used in `biom/table.py` and `biom/tests/test_table.py` (`grep -ir skbio`).

> I'd suggest/commit but gh isn't letting me do so in the PR

Weird, "Allow edits by maintainers" is enabled in the PR.

wasade:
Clearly both my gh and grep foo are failing. 

it's an optional dependency, and only needed for doc build which is done via py36. Removed, let's see what happens

wasade:
Seems like some other dependency, not skbio, is being problematic:

https://github.com/biocore/biom-format/runs/4255448804?check_suite_focus=true

nsoranzo:
You are currently running flake8 twice in the GitHub workflow, once in the `lint` job and once (per Python version) in the `build` job.

wasade:
Could be, this was the first project we migrated to gh-actions from travis, and the config is still a little rough around the edges

wasade:
I'm going to restore the dependencies for right now, it doesn't seem like an skbio issue but rather somewhere else. I'll issue a separate PR to remove six/future as they're no longer necessary

nsoranzo:
> I'll issue a separate PR to remove six/future as they're no longer necessary

I've now removed those here, I think it's a good fit for this PR :)
Also, I've specified that conda should use conda-forge, I don't think `default` has so many Python 3.10 packages.

wasade:
You're quickly becoming my favorite person

wasade:
Unfortunately, those are real. Possible the h5py build in conda-forge is unusual?

nsoranzo:
> Unfortunately, those are real. Possible the h5py build in conda-forge is unusual?

@wasade The conda-forge build is not unusual, but when using conda-forge h5py 3.1.0 gets installed instead of 2.10, and this caused the issues. I've fixed them in 2 new commits, hope they make sense.

nsoranzo:
Looks like `nose` is needed for `@npt.dec.skipif()`, should that be replaced by the pytest equivalent?

wasade:
This is hitting some scope creep :/ if you're able that would be nice, thanks. No preference for pytest vs unittest.skipIf

nsoranzo:
All green :tada:

wasade:
Fantastic!!! Thank you so much! I'll look at a release in the coming weeks so this can be included quickly :)

Vlasovets:
hi @nsoranzo!

I'm experiencing an installation problem using Docker and Python 3.10. 
I was able to successfully install biom-format using on Google Colab using pip and python 3.10.11, but with Docker it surprisingly failed, may I ask you to navigate to a solution?

I attach the output describing the error (if it is ok), I could also paste it here, but it is rather long.
Otherwise, it is possible to reproduce the error, running the building the following Docker image:

`# syntax=docker/dockerfile:1
FROM quay.io/qiime2/core:2023.2
RUN apt-get update && apt-get -y install python3.10
WORKDIR /app
COPY requirements.txt . ### biom-format
RUN python3.10 -m pip install -r requirements.txt`

Thank you in advance!
[error.txt](https://github.com/biocore/biom-format/files/11376938/error.txt)

wasade:
Hi @Vlasovets, it looks like the environment is missing a compiler, see `error: command 'gcc' failed: No such file or directory` near the end of the error output

PR #866, opened 2021-11-19
Improve Table.collapse by roughly 10x

wasade:
For collapses requiring high volumes of partitions, these changes empirically offered about a 10x reduction in runtime. In particular, this greatly improves performance for `redbiom.fetch` when merging samples.


antgonza:
Looks great, thank you. Those improvements sounds exciting!

wasade:
thanks!

PR #881, opened 2022-09-20
take column names from all metadata not just first entry

asafpr:
When converting observation metadata to dataframe the column names were taken from the first row. When the first row had missing values (i.e. incomplete taxonomic assignment) then the number of columns would not match the column headers. I changed it to take the maximal list of column names from all rows.


wasade:
Hi @asafpr, sorry for the delay on this -- reviewing now

wasade:
Thanks!!!

PR #888, opened 2022-12-29
Avoid dense transform collapse

### LINKED ISSUE #887: Table.collapse performance issue
I think that the [`Table.collapse`](http://biom-format.org/documentation/generated/biom.table.Table.collapse.html) method may not have the best performance in its current form. Woltka calls this method to collapse BIOM tables. I found that it is quite slow when working with large BIOM tables, especially those of functional units (ORFs, KOs, UniRefs, etc.), which are common in shotgun metagenomics. For example, collapsing 2,065,892 ORFs by 25 samples (a very small test dataset) took 14m03.86s and 7.45 GB RAM. When working with real datasets it could be slower.

I did a simple benchmark using the [Moving Pictures](https://docs.qiime2.org/2022.8/tutorials/moving-pictures/) dataset, which has 545 ASVs mapped to 197 genera, from 34 samples. Using biom-format's `collapse`, it took **217 ms ± 838 µs**.

```python
res = table.collapse(lambda id_, _: map_[id_], norm=False, axis='observation')
```

I did some research and found that there may be more efficient solutions, at least in certain scenarios. The most straight-forward method may be Pandas' [`groupby`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html), which is well-optimized and generic. In my benchmark it took **1.37 ms ± 23.8 µs**.

```python
res = table.to_dataframe(dense=True).groupby(map_).agg(sum)
```

I found several articles discussing the underlying mechanisms and alternative implementations of `groupby`: [here](https://jakevdp.github.io/blog/2017/03/22/group-by-from-scratch/), [here](https://nicholasvadivelu.com/2021/02/16/groupby-from-scratch/), and [here](https://blogs.avsolatorio.com/optimizing-pandas-groupby-50x-using-numpy-to-speedup-an-agent-based-model-a-story-of-8b0d25614915). By learning from these articles, I also came up with a pure NumPy solution, which isn't as fast as `grouby` but close. It took **2.38 ms ± 2.7 µs**.

```python
keys = [map_[x] for x in table.ids('observation')]
uniq, group = np.unique(keys, return_inverse=True)
res = np.zeros([uniq.shape[0], table.shape[1]])
np.add.at(res, group, table.matrix_data.toarray())
```

Both the Pandas and the NumPy solutions may suffer from a heavy bottleneck of converting the sparse matrix in a BIOM table into a dense matrix. However, according to these articles, there appears to be a solution utilizing SciPy's sparse matrix. I don't quite understand it, but I guess there may be a solution that does the job inside BIOM without converting.

Attached is my code for benchmarking. @wasade 

[collapse.ipynb.zip](https://github.com/biocore/biom-format/files/10297509/collapse.ipynb.zip)

wasade:
We were accidentally performing a dense conversion during one-to-many collapse. This conversion is not necessary. 

Fixes #887


wasade:
Thanks!

PR #889, opened 2023-01-04
Remove duplicated install requirement

nsoranzo:


wasade:
Ah, wonderful catch, thank you!

PR #904, opened 2023-03-10
dict is sorted in currently supported versions of Python

peterjc:
No need to explicitly use OrderedDict anymore, originally added in #788 and version 2.1.7 when Python 2.7 was still supported.


wasade:
Thanks. And since we no longer support older versions we should be safe here.

PR #906, opened 2023-03-12
chore: remove cython from requirements

### LINKED ISSUE #898: pip check reports cython as a requirement
I have a [bioconda package](https://github.com/bioconda/bioconda-recipes/pull/39809) that depends on biom-format. Among its tests is a `pip check` which reports cython as missing.

I see two possible solution:

1. Add cython to the [conda-forge biom-format](https://github.com/conda-forge/biom-format-feedstock) feedstock as a run dependency. I think, this is the worse solution.
2. Remove cython from the `install_requires`. You already list it in `pyproject.toml` and cython is imported by `setup.py` itself. So at that point it's too late to install the requirement anyway. I consider this the better option. Unless, of course, cython is truly needed at runtime but I don't see why that would be the case.

Midnighter:
I didn't see instructions regarding whether the PR should be merged into master or develop.

Fix #898 

Remove Cython from the package requirements since it's only needed at installation time and already listed in `pyproject.toml` build requirements.

I didn't make any other changes here but I did notice that the whole package installation/building configuration uses outdated mechanisms. You may consider making further changes:

* Setuptools now [fully supports pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). This would let you get rid of `setup.cfg` and `pytest.ini` as well.
* Supporting `python setup.py test` is not typically done anymore.
* You can pre-compile your `.pyx` modules into `.c` ones and make those part of your source distribution instead. In that case, cython wouldn't even be needed at installation time but only when making your `sdist`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -116,7 +116,6 @@ install_requires = [
     "numpy >= 1.9.2",
     "scipy >= 1.3.1",
     'pandas >= 0.20.0',
-    "cython >= 0.29",
     "h5py",
 ]
 


wasade:
Thanks, @Midnighter! And thank you for the additional notes on items that we can be improved! I usually only get brief windows of time for maintenance once or twice a year, and I've preferentially focused on bugs / functionality / performance. That said, I would like to spend a cycle at some point on project-level maintenance.

Midnighter:
That's... not very often 😅

wasade:
Ya :) but the items you raise are important and I would like to make sure we keep forward progress (even if slow) rather than stalling out.

PR #916, opened 2023-03-24
Explicitly allow setting a random seed for subsample

### LINKED ISSUE #914: Setting a random seed
Documentation and a verifiable example of setting a random seed should be provided.

wasade:
Fixes #914. You can now specify the random seed on call to `Table.subsample(...)`.

cc @gibsramen @rob-knight

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,7 +4,9 @@ BIOM-Format ChangeLog
 biom 2.1.14-dev
 ---------------
 
-Changes go here
+New features:
+
+* You can now set a random seed at the call to `Table.subsample`, see issue [#914](https://github.com/biocore/biom-format/issues/914).
 
 biom 2.1.14
 -----------
--- a/biom/_subsample.pyx
+++ b/biom/_subsample.pyx
@@ -11,7 +11,7 @@ import numpy as np
 cimport numpy as cnp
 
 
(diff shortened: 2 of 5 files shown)


gibsramen:
Thanks, @wasade. How much work would it be to rewrite the random shuffling into the new NumPy random API? Setting seed with `np.random.seed` is now a [legacy function](https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html) that sets the *global* random seed which can have some unintended consequences. The new RNG documentation is [here](https://numpy.org/doc/stable/reference/random/index.html#random-sampling-numpy-random).

wasade:
Good observation. Likely easy given the methods are there but the signature for the underlying cython method will need a minor adjustment

```
>>> from numpy.random import default_rng
>>> rng = default_rng(12345)
>>> rng.multinomial
<built-in method multinomial of numpy.random._generator.Generator object at 0x7f9c385c1900>
>>> rng.permutation
<built-in method permutation of numpy.random._generator.Generator object at 0x7f9c385c1900>
>>> 
```

wasade:
@gibsramen if this is green can you merge?

gibsramen:
Looks good, thanks! I think this is good to merge.

wasade:
Thanks!

PR #919, opened 2023-04-04
Disallow head to have values <= 0

### LINKED ISSUE #899: biom head -n 0 and -m 0 not supported
I consider this low priority, but probably an easy fix.

# Observed behaviour

Attempting to use ``biom head -n 0 ...`` and/or ``-m 0`` both fail with an ugly traceback ending ``IndexError: n cannot be <= 0.`` or ``IndexError: m cannot be <= 0.``

Tested using macOS, biom-format installed via conda:

```
$ biom --version
biom, version 2.1.14
```

# Expected behaviour:

Using ``-m 0`` would show just the first column of identifiers (and no columns of samples). Work around:

``$ biom head -m 1 example.biom | cut -f 1``

Using ``-n 0`` would show just the headers (including sample names unless ``-m 0``), but none of the observations. Workaround:

``$ biom head -n 1 example.biom | grep "^#"``

wasade:
Fixes #899 

@peterjc, I'm not allowing values of 0 here as obtaining sample or observation IDs can be performed with `biom table-ids`

### DIFF
--- a/biom/cli/table_head.py
+++ b/biom/cli/table_head.py
@@ -36,6 +36,12 @@ def head(input_fp, output_fp, n_obs, n_samp):
     $ biom head -i table.biom
 
     """
+    if n_obs <= 0:
+        raise ValueError("-n/--n-obs must be > 0")
+
+    if n_samp <= 0:
+        raise ValueError("-m/--m-samp must be > 0")
+
     table = load_table(input_fp).head(n=n_obs, m=n_samp)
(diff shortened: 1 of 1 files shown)


peterjc:
Perhaps special case 0 with a recommendation to use ``biom table-ids``, with negative numbers giving the error as written?

peterjc:
Looks good to me :)

PR #920, opened 2023-04-05
Issue 896

wasade:
Remove h5py optional considerations

### DIFF
--- a/biom/cli/metadata_adder.py
+++ b/biom/cli/metadata_adder.py
@@ -13,7 +13,6 @@ from biom import load_table
 from biom.cli import cli
 from biom.cli.util import write_biom_table
 from biom.parse import MetadataMap
-from biom.util import HAVE_H5PY
 
 
 @cli.command(name='add-metadata')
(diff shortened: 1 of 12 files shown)


peterjc:
Minor comment: I'd put the ``import h5py`` line before your local imports. I can recommend tools to do this sorting automatically.

Flake8 still failing with several ``F401 'pytest' imported but unused``

peterjc:
Looks like where you've removed the ``@pytest.mark.skipif`` decorators, some of those files now have a redundant ``import pytest`` which is why flake8 complains.

wasade:
Missing flake8 is what I get for pushing in changes at the end of the day... but it's also a perpetual reminder of the benefit of switching to black.

peterjc:
I know what you mean. Note black wouldn't have caught this though.

Personally I now use the tool pre-commit framework to run flake8 and black as a git pre-commit hook (and as part of CI testing), https://pre-commit.com/

wasade:
That's a great find, thank you!

PR #934, opened 2023-06-12
Implement subsample_without_replacement using C++17 std::sample

sfiligoi:


sfiligoi:
CC @wasade

sfiligoi:
This version uses drastically less memory, as it does not expand the buffer using np.repeat.
Should be cheaper compute wise, too.

sfiligoi:
Alternative version available in #935 .

PR #935, opened 2023-06-13
Use rng.choice without unpacked in subsample_without_replacement with 64-bit support

sfiligoi:


sfiligoi:
Drastically reduces the memory needs when sums are large.

biom/_subsample.pyx
```diff
@@ -39,33 +35,110 @@ def _subsample(arr, n, with_replacement, rng):
...
+
+    Notes
+    -----
+    This code was adapted from scikit-bio (`skbio.math._subsample`)
+
+    """
+    cdef:
+        cnp.int64_t counts_sum,idx,count_el, perm_count_ela
+        cnp.int64_t count_rem
+        cnp.int64_t cn = n
```
wasade, line 76: Shouldn't be necessary if using `cdef` on the function and typing the arguments

biom/_subsample.pyx
```diff
@@ -39,33 +35,110 @@ def _subsample(arr, n, with_replacement, rng):
...
-            result = np.zeros(length, dtype=np.float64)
-            for idx in range(permuted.shape[0]):
-                result[permuted[idx]] += 1
-
-            data[start:end] = result
+        if counts_sum < cn:
+            data[start:end] = 0
+            continue
+
+        permuted = rng.choice(counts_sum,cn,replace=False,shuffle=False)
```
wasade, line 93:
```suggestion
        permuted = rng.choice(counts_sum, cn, replace=False, shuffle=False)
```

biom/_subsample.pyx
```diff
@@ -39,33 +35,110 @@ def _subsample(arr, n, with_replacement, rng):
...
+        permuted.sort()
+
+        # now need to do reverse mapping
+        result = np.zeros(length, dtype=np.float64)
+        el=0         # index in result/data
+        count_el =0  # index in permutted
+        count_rem=data[start]  # since each data has multiple els, sub count there
```
wasade, line 100:
```suggestion
        el = 0         # index in result/data
        count_el = 0  # index in permutted
        count_rem = data[start]  # since each data has multiple els, sub count there
```

biom/_subsample.pyx
```diff
@@ -39,33 +35,110 @@ def _subsample(arr, n, with_replacement, rng):
...
+        for idx in range(cn):
```
wasade, line 101: Can `idx` be declared? It may help to run `cython -a` on this file, the annotation report will flag high code generation lines and I think this will happen here as `idx` doesnt' look typed

biom/_subsample.pyx
```diff
@@ -39,33 +35,110 @@ def _subsample(arr, n, with_replacement, rng):
...
+            perm_count_el = permuted[idx]
+            # the array is sorted, so just jump ahead
+            while (perm_count_el-count_el) >= count_rem:
+               count_el += count_rem
+               el += 1
+               count_rem = data[start+el]
```
wasade, line 107:
```suggestion
            while (perm_count_el - count_el) >= count_rem:
               count_el += count_rem
               el += 1
               count_rem = data[start+el]
```

biom/_subsample.pyx
```diff
@@ -39,33 +35,110 @@ def _subsample(arr, n, with_replacement, rng):
...
+            count_rem -= (perm_count_el-count_el)
```
wasade, line 108:
```suggestion
            count_rem -= (perm_count_el - count_el)
```

biom/_subsample.pyx
```diff
@@ -39,33 +35,110 @@ def _subsample(arr, n, with_replacement, rng):
...
+    ndarray
+        Subsampled data
+
+    Notes
+    -----
+    This code was adapted from scikit-bio (`skbio.math._subsample`)
+
+    """
+    cdef:
+        cnp.int64_t counts_sum,idx,count_el, perm_count_ela
```
wasade, line 74:
```suggestion
        cnp.int64_t counts_sum, idx, count_el, perm_count_ela
```

biom/_subsample.pyx
```diff
@@ -39,33 +35,110 @@ def _subsample(arr, n, with_replacement, rng):
...
+        permuted = rng.choice(counts_sum,cn,replace=False,shuffle=False)
+        permuted.sort()
+
+        # now need to do reverse mapping
+        result = np.zeros(length, dtype=np.float64)
+        el=0         # index in result/data
+        count_el =0  # index in permutted
+        count_rem=data[start]  # since each data has multiple els, sub count there
+        for idx in range(cn):
+            perm_count_el = permuted[idx]
```
wasade, line 102: I think this is typed as `perm_count_ela`

biom/_subsample.pyx
wasade: Sorry I see it now. I think it may trigger a cast if the idx is not pyssize_t

wasade (changes requested):

biom/_subsample.pyx
sfiligoi: Yep. Fixed.

biom/_subsample.pyx
sfiligoi: Done

biom/_subsample.pyx
```diff
@@ -6,22 +6,18 @@
...
 # -----------------------------------------------------------------------------
 
-
 import numpy as np
 cimport numpy as cnp
 
-
-def _subsample(arr, n, with_replacement, rng):
-    """Subsample non-zero values of a sparse array
+def _subsample_with_replacement(arr, n, rng):
```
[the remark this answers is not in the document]
sfiligoi: I was trying to stay as close as possible to the original code, but it is a good idea.
Done.

biom/_subsample.pyx
```diff
@@ -38,34 +39,119 @@ def _subsample(arr, n, with_replacement, rng):
...
+
+        permuted = rng.choice(counts_sum, n, replace=False, shuffle=False)
+        permuted.sort()
+
+        # now need to do reverse mapping
+        # since I am not using np.repeat anymore
+        # reminder, old logic was
+        #   r = np.arange(length)
+        #   unpacked = np.repeat(r, data_i[start:end])
+        #   permuted_unpacked = rng.choice(unpacked, n, replace=False, shuffle=False)
```
wasade, line 107:
```suggestion
        #   permuted_unpacked = rng.choice(unpacked, n, replace=False, shuffle=False)
        # 
        # specifically, what we're going to do here is randomly pick what elements within
        # each sample to keep. this is analogous issuing the prior np.repeat call, and obtaining
        # a random set of index positions for that resulting array. however, we do not need to 
        # perform the np.repeat call as we know the length of that resulting vector already,
        # and additionally, we can compute the sample associated with an index in that array
        # without constructing it.
```

Is the above correct and make sense?

wasade (changes requested):
Could a note be added to the change log with approximate performance boost and memory reduction?

biom/_subsample.pyx
sfiligoi: Yes, it is correct.

sfiligoi:
On EMP-style BIOM, the new without_replacement algorithm is about 2x faster (on a EPYC 7302 CPU):
33s vs 58s

There is also a small reduction is memory consumption, but it is barely noticeable compared to the rest of the memory use when using the Table object.

sfiligoi:
For BIOM tables with very large per-column sums, it is an enabler;
the old code would outright fail if the columns sum was > 2^31.
(e.g. rna_integrity_metaG_woltka_wolr2_with_plasmids_per_gene_clean_gene_counts_per_g_stool.biom)

(But  just fixing that would not help... a test showed it would have needed over 15.1 TiB of RAM) 

The new code has a 2^63 limit, and the memory use is proportional to n, only.

The running time is quite fast for small n, but gets slower with high n (on same CPU as above):
n=2000 50s
n=6G 20min

wasade (approved):

wasade:
Thanks!!

PR #938, opened 2023-09-13
More gracefully deal with not-perfectly-integer inputs

sfiligoi:
The input data is supposed to be integer, but it is represented using fp64.
If the elements are not precisely integer, the previous logic would fail, possibly leading to data corruption.
The new version enforces the integer constraint.


wasade:
Thanks!! Could a note be added to the change log?

wasade:
Great, thank you!!!

PR #949, opened 2024-04-04
WIP Supporting Windows and Python 3.12

qiyunzhu:
This PR adds build support for Windows and Python 3.12. Following a discussion with @wasade and @mataton, we realized that biom-format might have already supported Windows, and it just needs to be officially added. The CI workflow will tell whether it is the case.

PR #951, opened 2024-04-05
Windows support

mataton:
A continuation of PR #949, this PR adds support for Windows.

All issues revolve around two things:
1) File path separators differing between operating systems. `/` vs. `\`
2) Temporary file handling differing between operating systems.

The first issue is solved with a straightforward fix using `os.join()` and `os.path.sep`.

The second issue is more complex. In Windows, the name of a `NamedTemporaryFile` object **cannot** be used to open the file a second time, while the named temporary file is still open. See this [SO](https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file), and the Python [docs](https://docs.python.org/3.9/library/tempfile.html#tempfile.NamedTemporaryFile). The strategy is to use the `delete=False` parameter to prevent the temporary file from being deleted when it is closed, and then manually delete the file when the test is done using it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ jobs:
     name: Build sdist
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Build distribution
         run: |
@@ -15,7 +15,7 @@ jobs:
           pip install numpy cython
           pipx run build --sdist
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: dist-artifacts
           path: dist/*.tar.gz
@@ -27,13 +27,13 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
-        pyver: ["37", "38", "39", "310", "311"]
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        pyver: ["37", "38", "39", "310", "311", "312"]
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: 3.9
 
@@ -47,7 +47,7 @@ jobs:
       - name: Build wheels (py ${{ matrix.pyver }}) Linux
         if: matrix.os == 'ubuntu-latest' 
         env:
-          CIBW_ARCHS_LINUX: x86_64
+          CIBW_ARCHS_LINUX: "x86_64 aarch64"
           CIBW_SKIP: "*-musllinux*"
           CIBW_BUILD: "cp${{ matrix.pyver }}-*"
 
@@ -60,10 +60,15 @@ jobs:
           CIBW_BUILD: "cp${{ matrix.pyver }}-*"
 
         uses: pypa/cibuildwheel@v2.12.3
-      
-      
+
+      - name: Build wheels (py ${{ matrix.pyver }}) Windows
+        if: matrix.os == 'windows-latest'
+        env:
+          CIBW_ARCHS_WINDOWS: "amd64 win32 arm64"
+          CIBW_BUILD: "cp${{ matrix.pyver }}-*"
+
       - name: Upload wheels
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: dist-artifacts
           path: ./wheelhouse/*.whl
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,7 @@ biom 2.1.15-dev
 
 Performance improvements:
 
+* Add Windows support. PR[#951](https://github.com/biocore/biom-format/pull/951) revises codebase to be Winows compatible and adds this support to the CI testing matrix.
 * Add NumPy 2.0 support. PR [#950](https://github.com/biocore/biom-format/pull/950) ensures code compatibility with NumPy 2.0. This support is yet to be added to the CI testing matrix.
 * Revise `Table._fast_merge` to use COO directly. For very large tables, this reduces runtime by ~50x and memory by ~5x. See PR [#913](https://github.com/biocore/biom-format/pull/933).
 * Drastically reduce the memory needs of subsampling when sums are large. Also adds 64-bit support. See PR [#935](https://github.com/biocore/biom-format/pull/935).
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,7 @@ Performance improvements:
 
 * Revise `Table._fast_merge` to use COO directly. For very large tables, this reduces runtime by ~50x and memory by ~5x. See PR [#913](https://github.com/biocore/biom-format/pull/933).
 * Drastically reduce the memory needs of subsampling when sums are large. Also adds 64-bit support. See PR [#935](https://github.com/biocore/biom-format/pull/935).
+* Improve handling of not-perfectly-integer inputs. See PR [#938](https://github.com/biocore/biom-format/pull/938).
 
 biom 2.1.15
 -----------
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,7 +6,7 @@ biom 2.1.15-dev
 
 Performance improvements:
 
-* Add Windows support. PR[#951](https://github.com/biocore/biom-format/pull/951) revises codebase to be Winows compatible and adds this support to the CI testing matrix.
+* Add Windows support. PR[#951](https://github.com/biocore/biom-format/pull/951) revises codebase to be Windows compatible and adds this support to the CI testing matrix.
 * Add NumPy 2.0 support. PR [#950](https://github.com/biocore/biom-format/pull/950) ensures code compatibility with NumPy 2.0. This support is yet to be added to the CI testing matrix.
 * Revise `Table._fast_merge` to use COO directly. For very large tables, this reduces runtime by ~50x and memory by ~5x. See PR [#913](https://github.com/biocore/biom-format/pull/933).
 * Drastically reduce the memory needs of subsampling when sums are large. Also adds 64-bit support. See PR [#935](https://github.com/biocore/biom-format/pull/935).
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,7 @@ biom 2.1.15-dev
 
 Performance improvements:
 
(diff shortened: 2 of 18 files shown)


biom/tests/test_cli/test_add_metadata.py
```diff
@@ -20,10 +21,11 @@ class TestAddMetadata(TestCase):
     def setUp(self):
         """Set up data for use in unit tests."""
         self.cmd = _add_metadata
-        with tempfile.NamedTemporaryFile('w') as fh:
+        with tempfile.NamedTemporaryFile('w', delete=False) as fh:
             fh.write(biom1)
             fh.flush()
             self.biom_table1 = biom.load_table(fh.name)
+        os.unlink(fh.name)
```
wasade, line 28: Could `self.temporary_fh_name` be set within the scope of the context manager, and move the `os.unlink` to `tearDown`? The reason for the former is the current approach relies on scope creep which works but borders on surprising behavior. The latter is important as, if a test fails, cleanup will only reliably occur if the `unlink` is performed by `tearDown`.

biom/tests/test_cli/test_summarize_table.py
```diff
@@ -12,16 +12,18 @@
...
 
 class TestSummarizeTable(TestCase):
 
     def setUp(self):
-        with tempfile.NamedTemporaryFile(mode='w') as fh:
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as fh:
             fh.write(biom1)
             fh.flush()
             self.biom1 = load_table(fh.name)
+        os.unlink(fh.name)
```
wasade, line 26: Same here per comment above

biom/tests/test_cli/test_table_converter.py
```diff
@@ -28,16 +29,19 @@ def setUp(self):
         self.cmd = _convert
         self.output_filepath = tempfile.NamedTemporaryFile().name
 
-        with tempfile.NamedTemporaryFile('w') as fh:
+        with tempfile.NamedTemporaryFile('w', delete=False) as fh:
             fh.write(biom1)
             fh.flush()
             self.biom_table1 = load_table(fh.name)
+            fh.close()
```
wasade, line 36: Was it necessary to `close` here but not in the other `TestCase`s? And same here re `unlink`ing. If it's viable to just unlink within the context manager than that may be easier than the `tearDown`

biom/tests/test_parse.py
```diff
@@ -281,38 +281,40 @@ def test_parse_adjacency_table_no_header(self):
...
 
     def test_save_table_filepath(self):
         t = Table(np.array([[0, 1, 2], [3, 4, 5]]), ['a', 'b'],
                   ['c', 'd', 'e'])
-        with NamedTemporaryFile() as tmpfile:
+        with NamedTemporaryFile(delete=False) as tmpfile:
             save_table(t, tmpfile.name)
             obs = load_table(tmpfile.name)
             self.assertEqual(obs, t)
+        os.unlink(tmpfile.name)
```
wasade, line 297: Same here

biom/tests/test_table.py
```diff
@@ -1016,64 +1016,72 @@ def test_to_from_hdf5_bug_861(self):
...
-        with NamedTemporaryFile() as tmpfile:
+        with NamedTemporaryFile(delete=False) as tmpfile:
             h5 = h5py.File(tmpfile.name, 'w')
             t.to_hdf5(h5, 'tests')
             h5.close()
 
             h5 = h5py.File(tmpfile.name, 'r')
             obs = Table.from_hdf5(h5)
+            h5.close()
+        os.unlink(tmpfile.name)
```
wasade, line 1027: same here

biom/tests/test_table.py
```diff
@@ -1016,64 +1016,72 @@ def test_to_from_hdf5_bug_861(self):
...
+        with NamedTemporaryFile(delete=False) as tmpfile:
             h5 = h5py.File(tmpfile.name, 'w')
             t.to_hdf5(h5, 'tests', creation_date=current)
             h5.close()
 
             h5 = h5py.File(tmpfile.name, 'r')
             obs = Table.from_hdf5(h5)
             self.assertEqual(obs.create_date, current)
+            h5.close()
+        os.unlink(tmpfile.name)
```
wasade, line 1044: same here

biom/tests/test_table.py
```diff
@@ -1016,64 +1016,72 @@ def test_to_from_hdf5_bug_861(self):
...
     def test_to_hdf5_empty_table(self):
         """Successfully writes an empty OTU table in HDF5 format"""
         # Create an empty OTU table
         t = Table([], [], [])
-        with NamedTemporaryFile() as tmpfile:
+        with NamedTemporaryFile(delete=False) as tmpfile:
             h5 = h5py.File(tmpfile.name, 'w')
             t.to_hdf5(h5, 'tests')
             h5.close()
+        os.unlink(tmpfile.name)
```
wasade, line 1056: same here

biom/tests/test_table.py
```diff
@@ -1016,64 +1016,72 @@ def test_to_from_hdf5_bug_861(self):
...
 
     def test_to_hdf5_empty_table_bug_619(self):
         """Successfully writes an empty OTU table in HDF5 format"""
         t = example_table.filter({}, axis='observation', inplace=False)
-        with NamedTemporaryFile() as tmpfile:
+        with NamedTemporaryFile(delete=False) as tmpfile:
             h5 = h5py.File(tmpfile.name, 'w')
             t.to_hdf5(h5, 'tests')
             h5.close()
+        os.unlink(tmpfile.name)
```
wasade, line 1065: same here

wasade: ...I'll stop flagging on these

biom/tests/test_table.py
```diff
@@ -1184,18 +1199,19 @@ def test_to_hdf5(self):
             self.assertIn('observation', h5)
             self.assertIn('sample', h5)
             self.assertEqual(sorted(h5.attrs.keys()), sorted(['id', 'type',
-                                                              'format-url',
-                                                              'format-version',
-                                                              'generated-by',
-                                                              'creation-date',
-                                                              'shape', 'nnz']))
+                                                            'format-url',
```
wasade, line 1202: The change in indent here looks incorrect as it should be aligned with the `[`?

wasade (changes requested):
Thanks, @mataton!!! This is super helpful, and glad to see the changes seem relatively minor! The primary comments are capturing the tmp filename for unlinking in tearDown

biom/tests/test_cli/test_table_converter.py
mataton: If you `close` the file before `unlink`ing within the context manager, then it works. Otherwise you get a `PermissionError` at the `unlink`ing step. Looking at the code now, I think the idea of setting `self.temporary_fh_name` in the context manager and putting the `unlink` in `tearDown` is the best solution.

biom/tests/test_table.py
mataton: These tests are creating the temporary files outside of a `setUp`/`tearDown` context, so the solution involving putting the `unlink` in `tearDown` won't work here. The reason the `close`s are necessary within the context manager is that without them `unlink` will throw ` PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\me\\AppData\\Local\\Temp\\tmpg4pk3ink'` . My understanding is that the context manager will close `tmpfile` when it is finished, but it won't close `h5`. Exactly why this is the case I'm unsure of.

mataton: I think this is the case because I set `delete=False` during the file creation, so a file will not be deleted when it is closed. But it seems on windows that file handles must be closed before the file may be deleted.

mataton: So this solution works because it closes all file handles, including `tmpfile` and `h5`, and then deletes the file using `unlink`. I'm not sure of a better solution for this particular case and the other similar cases.

mataton:
@wasade Essentially it looks like there are two scenarios. One where all the tests in whichever `test.py` file require the temporary files. In this case, the `setUp`/`tearDown` solution I feel is best. This is true in `test_add_metadata.py`, `test_summarize_table.py`, and `test_table_converter.py`.

The other scenario is that not all of the tests within the `test.py` require the temporary files, so using `setUp`/`tearDown` framework doesn't make sense here. The current solution in those files I think might be the most appropriate, considering my previous comments.

.github/workflows/release.yml
```diff
@@ -60,10 +60,15 @@ jobs:
           CIBW_BUILD: "cp${{ matrix.pyver }}-*"
 
         uses: pypa/cibuildwheel@v2.12.3
-      
-      
+
+      - name: Build wheels (py ${{ matrix.pyver }}) Windows
+        if: matrix.os == 'windows-latest'
+        env:
+          CIBW_ARCHS_WINDOWS: "amd64 win32 arm64"
```
wasade, line 67: Is that 32-bit windows? If so, does the build actually work?

wasade: ...this repository does not need to support 32-bit architectures

biom/tests/test_table.py
wasade: Issue is edge case if test excepts prior to `unlink`. A common strategy for this type of stuff:

```python
def setUp(self):
    self.temp_files = []
def tearDown(self):
    for f in self.temp_files:
        os.unlink(f)
def test_some_test(self):
    ...
    self.temp_files.append(blah)
    ...
```

wasade:
Thanks, @mataton. I still think we do want `unlink` in `tearDown` as that ensures it runs even in the event of an exception, does that make sense? It's relatively minor here though as these are still going into `TMPDIR`

mataton:
@wasade Your suggested changes make sense, and there was actually already a `to_remove = []` in the `setUp` functions already that I didn't notice previously, which I used. Just had to add the proper `tearDown` to a couple files. It should be good now.

wasade (approved):

wasade:
Will merge once aarch64 completes

wasade:
...thank you @mataton, this is super helpful!!! One last request, could you add a changelog note?

mataton:
Happy to contribute!

ChangeLog.md
```diff
@@ -6,6 +6,7 @@ biom 2.1.15-dev
 
 Performance improvements:
 
+* Add Windows support. PR[#951](https://github.com/biocore/biom-format/pull/951) revises codebase to be Winows compatible and adds this support to the CI testing matrix.
```
wasade, line 9:
```suggestion
* Add Windows support. PR[#951](https://github.com/biocore/biom-format/pull/951) revises codebase to be Windows compatible and adds this support to the CI testing matrix.
```

PR #953, opened 2024-04-10
Removed dependency on natsort

qiyunzhu:
The package `natsort` is listed as a dependency but it is actually never used. Instead, BIOM has a built-in `natsort` function. Therefore this dependency can be removed. See:

https://github.com/scikit-bio/scikit-bio/issues/2007


wasade:
Am +1 assuming all green

qiyunzhu:
@wasade All greened.

PR #961, opened 2024-05-02
Issue 952

### LINKED ISSUE #952: Edge case for biom.Table subsample with replacement
This was originally posted on the qiime2 issue tracker https://github.com/qiime2/q2-feature-table/issues/245

I'm able to replicate it with some of my own data, and the results are a bit perplexing. When I pull out the original cython code and run it directly in ipython, it actually works -- suggesting to me that there is something off with the compilation of the cython code.

Below is the code to reproduce the error when I run `table.subsample`
```python
import biom
table= biom.load_table('table.biom')
table.subsample(10, axis='sample', by_id=False, with_replacement=True)
```
And below is the code to directly compute the subsampling (borrowed from [here](https://github.com/biocore/biom-format/blob/master/biom/_subsample.pyx#L12-L52)), which doesn't have an error

```python
import biom
table= biom.load_table('table.biom')
arr = table._get_sparse_data()

data, indptr = arr.data, arr.indptr
for i in range(indptr.shape[0] - 1):
        start, end = indptr[i], indptr[i+1]
         length = end - start
         counts_sum = data[start:end].sum()
         pvals = data[start:end] / counts_sum

         data[start:end] = rng.multinomial(n, pvals)
```
See attached zip file [table.zip](https://github.com/biocore/biom-format/files/14907539/table.zip) to reproduce this error. This was reproduced in biom versions 2.1.15, as well as 2.1.14 and 2.1.11

I'm also not seeing this issue when I'm using the dense subsampling in skbio
```python
import biom
from skbio.stats import subsample_counts
table= biom.load_table('table.biom')
df = table.to_dataframe().T
df.apply(lambda x: subsample_counts(x.astype(np.int64), 1e6, replace=True))
```

EDIT : It looks like there is a type mis-alignment regarding the types. If I cast everything to ints, the issue appears to be resolved
```python
import biom
table= biom.load_table('table.biom')
df = table.to_dataframe().apply(np.ceil).apply(np.int64).T
t = biom.Table(df.T.values, df.columns, df.index)
t.subsample(10, axis='sample', by_id=False, with_replacement=True)
```

wasade:
Fixes #952 
Fixes https://github.com/qiime2/q2-feature-table/issues/245

@mortonjt, are you able to review?


mortonjt:
Hi, I think this fix is good enough.  We just need to keep a mental note that `np.ceil` is applied to the data before subsampling to ensure that all low-abundance datasets have a chance to be sampled.  But given that this is a nuanced technical detail, I'm OK with merging as is.

wasade:
Wonderful, thank you! And good point -- I'll add a mention in the docstring.

wasade:
@mortonjt, please let me know if there are concerns with that text otherwise I'll merge

mortonjt:
LGTM

PR #975, opened 2024-08-04
chore: ensure that package is built against numpy >=2

### LINKED ISSUE #974: Numpy compatibility issue
When I install biom-format and numpy >= 2 in an environment, I get a lot of errors of the sort:

```
ValueError('numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject')
```

That is even though, version 2.1.16 promises numpy 2.0 compatibility.

I think, the reason is that you do not restrict the numpy version in your build instructions and your wheels are compiled against numpy <2.

```toml
[build-system]
requires = ["setuptools","wheel", "numpy", "cython"]
```

Midnighter:
Fix #974

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools","wheel", "numpy", "cython"]
+requires = ["setuptools","wheel", "numpy ~=2.0", "cython"]


Midnighter:
Oh yeah, numpy has dropped support for Python <3.9...

Midnighter:
Seems like more of the package infrastructure is not up to date.

```
RuntimeError: Support for the test command was removed in Setuptools 72
```

PR #985, opened 2024-12-02
Remove `setuptools.command.test` as it is no longer supported.

mataton:
Support for the `test` command was [removed](https://github.com/pypa/setuptools/issues/4519) in setuptools 72. Currently, BIOM relies on this command to manage unit tests. However, if one runs `make test` on the current codebase, you'll see `RuntimeError: Support for the test command was removed in Setuptools 72`. This PR is a simple implementation to remove BIOM's dependency on this unsupported function. However, it may not be the ideal solution. More seasoned members of the BIOM dev team, please feel free to review and critique this approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,9 @@
 # ----------------------------------------------------------------------------
 
 ifeq ($(WITH_DOCTEST), TRUE)
-	TEST_COMMAND = python setup.py test -a --doctest-modules --doctest-glob='*.pyx'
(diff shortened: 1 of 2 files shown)


mataton:
Additional info [here](https://docs.pytest.org/en/stable/explanation/goodpractices.html#do-not-run-via-setuptools).

wasade:
Thanks, it looks like this may avoid Cython compilation?

wasade:
Merged into #986

Issue #984, opened 2024-11-05
Reduce import time


wasade:
See `python -X importtime -c "import biom`

PR #987, opened 2024-12-03
Speedup import time with lazy imports

mataton:
Addresses #984 

The average import time (`python -X importtime -c 'import biom'`) over 7 runs for the current master branch of BIOM is 0.44s. Moving the imports of the required dependencies into their respective functions brings the average import time down to 0.19s, a **2.3x** speedup.

Performance of individual functions is only impacted on the first call. For example, the first call to `Table.to_dataframe()` is 247 microseconds, while all subsequent calls are 180 microseconds. This is due to the module being cached in `sys.modules` after the first call. The execution time of `to_dataframe()` without the import statement inside the function is 180 microseconds.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,10 @@ BIOM-Format ChangeLog
 biom-2.1.16-dev
 ---------------
 
+Performance improvements:
+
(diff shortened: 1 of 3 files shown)


wasade:
Addresses #984

wasade:
Thanks @mataton! Would you be able to add a brief note to the change log? I'll merge after

mataton:
@wasade Thank you for the reminder!

PR #988, opened 2024-12-03
Support for almost equal like equality 

### LINKED ISSUE #983: Support `assertAlmostEqual()`
When testing, biom tables can be passed to an `assertEqual()` call, but it looks like they don't support `assertAlmostEqual()`.  It would be a great addition if there was a standard, built-in way for checking if two tables are almost equal (e.g. to within some tolerance).

wasade:
Fixes #983, partially addresses #983.

Note that error checking for `nan` is coming in a separate PR. 

cc @AmandaBirmingham

### DIFF
--- a/biom/table.py
+++ b/biom/table.py
@@ -1864,6 +1864,10 @@ class Table:
         allclose_kwargs : dict
             Any keyword arguments to provide to np.allclose
 
+        Notes
+        -----
+        Specify `equal_nan=True` to allow Nan to test equal.
+
         Returns
         -------
         bool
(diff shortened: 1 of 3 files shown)


biom/table.py
```diff
@@ -1879,19 +1927,10 @@ def _data_equality(self, other):
         necessary before performing the final comparison.
```
AmandaBirmingham, line 1927: Are all these docstring comments ("Matrices are equal iff the following items are equal <etc>") still appropriate to this particular method?  I'm wondering if they would make more sense on `__eq__` or with some of them going onto `_data_equality_meta`.

AmandaBirmingham (commented):
yay!

biom/table.py
wasade: This particular method, `_data_equality` is called on `__eq__` and should not exhibit different semantics from prior versions. However, I agree that they could be shuffled to more precise locations -- thanks! I'll do that in a moment

PR #989, opened 2024-12-03
Error handling

### LINKED ISSUE #982: Throw error if trying to create a biom table containing NaNs
Currently, the code allows a user to create a biom table that contains NaNs.  However, trying to use such a table frequently leads to obscure errors: the equality test doesn't work (because NaN != NaN), transform doesn't work for sparse representations (because 0/NaN is set to 0, *not* NaN, because transform doesn't touch zero entries in sparse matrix), and sometimes even doing an explicit NaN-aware compare (where you check whether the same positions contain NaN in two tables) fails (for a non-NaN-related reason :) because `matrix.data`'s order is different depending on whether it was instantiated as sparse or not. (@wasade indicates that the usual way to handle NaNs in a biom table is to set them to zero, which makes these issues go away.)

Since NaNs can't be truly supported in the current code, as a user, I would like it a lot if I just got an error **up front** if I tried to make a biom table that contained a NaN, rather than hitting an inscrutable issue downstream.

wasade:
Fixes #982 by allowing a user to specify `hasnan` as a raisable state, e.g.:

```python
with errstate(hasnan='raise'):
    table = Table(...)
```

The default behavior is to ignore as that was the prior implicit behavior. 

Support for `hasinf` was also added. This PR includes also includes some normalization of strings in `err.py` and an expanded set of tests to assert raise behaviors.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,7 @@ biom-2.1.16-dev
 
 Maintenance:
 
+* Allow testing for nan and inf values with "hasnan" and "hasinf" error states. See issue [#982](https://github.com/biocore/biom-format/issues/982).
 * Python 3.7 and 3.8 removed from CI as they are [end-of-life](https://devguide.python.org/versions/). Python 3.13 added to CI. See PR[#986](https://github.com/biocore/biom-format/pull/986).
 
 biom 2.1.16
--- a/biom/err.py
+++ b/biom/err.py
@@ -31,6 +31,12 @@ sampmdsize : 'raise'
     Treatment of a table in which the number of sample metadata elements
     differs from the size of the data.
 
+hasnan : 'ignore'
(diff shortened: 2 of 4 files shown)


AmandaBirmingham (commented):
Nice!  I was unaware of the `with errstate` functionality; is there documentation for it somewhere?  I couldn't find it in this repo or on biom-format.org.

wasade:
It's documented [here](https://github.com/biocore/biom-format/blob/master/biom/err.py#L3). I'm unsure at the moment why it doesn't appear on the website. To be honest, I think much of the documentation system may have been setup by @gregcaporaso 12-13 years ago, and it could certainly use a refresh. The approach @qiyunzhu has taken with scikit-bio is quite pleasing. @AmandaBirmingham, do you by chance have interest in examining whether some small changes in the documentation system may be viable?

PR #1000, opened 2025-12-22
Utf8 regression

wasade:
A regression on parse of non-trivial UTF-8 characters was introduced. Reported [here](https://forum.qiime2.org/t/feature-table-gives-ascii-codec-cant-decode-byte-0xc3-in-position-4-error/33858/7).

### DIFF
--- a/biom/table.py
+++ b/biom/table.py
@@ -4182,11 +4182,15 @@ html
         def axis_load(grp):
             """Loads all the data of the given group"""
             # fetch all of the IDs
-            ids = grp['ids'].asstr()[:]
+            ids = grp['ids']
 
             if ids.size > 0:
+                ids = ids.asstr()[:]
                 ids_dtype = 'U%d' % max([len(v) for v in ids])
                 ids = np.asarray(ids, dtype=ids_dtype)
+            else:
+                # .asstr does not handle an empty dataset
+                ids = ids[:]
 
             parser = defaultdict(lambda: general_parser)
             parser['taxonomy'] = vlen_list_of_str_parser
--- a/biom/table.py
+++ b/biom/table.py
@@ -4182,7 +4182,7 @@ html
         def axis_load(grp):
             """Loads all the data of the given group"""
             # fetch all of the IDs
-            ids = grp['ids'][:]
+            ids = grp['ids'].asstr()[:]
 
             if ids.size > 0:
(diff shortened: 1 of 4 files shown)


biom/tests/test_parse.py
```diff
@@ -283,6 +284,15 @@ def test_parse_adjacency_table_no_header(self):
         obs = Table.from_adjacency(''.join(lines))
         self.assertEqual(obs, exp)
 
+    def test_parse_biom_table_hdf5_accent_utf8_regression(self):
+        tab = example_table.copy()
+        tab.update_ids({i: f'{i}fóó' for i in tab.ids()}, inplace=True)
+        with NamedTemporaryFile(delete=False) as tmpfile:
```
ebolyen, line 290: Should `delete` be `False`? I think this would leave behind a file on the system.

biom/table.py
```diff
@@ -4182,11 +4182,15 @@ def ensure_utf8(x):
...
             # fetch all of the IDs
-            ids = grp['ids'][:]
+            ids = grp['ids']
 
             if ids.size > 0:
+                ids = ids.asstr()[:]
                 ids_dtype = 'U%d' % max([len(v) for v in ids])
                 ids = np.asarray(ids, dtype=ids_dtype)
+            else:
+                # .asstr does not handle an empty dataset
```
ebolyen, line 4192: 🫤 that feels rude of h5py

ebolyen (approved):
LGTM, thanks @wasade! One minor question on the test setup.

biom/tests/test_parse.py
wasade: Potentially, but err'ing on following the style of the rest of the script. And, a left over file in `TMPDIR` isn't the best but not the worst thing to do

biom/table.py
wasade: Might be HDF5 as the C-API is touchy. Or, we may not be correctly enforcing types on empty tables

biom/tests/test_parse.py
ebolyen: Oh I see. It looks like those add this to the end of the test:

```python
self.to_remove.append(tmpfile.name)
```

It's mostly github's problem these days I suppose. So your call, but wanted to mention it.

PR #1004, opened 2026-05-26
upgrade cibuildwheel action to support Python 3.14

### LINKED ISSUE #1002: Outdated action pypa/cibuildwheel
Hello,

You are using a very outdated version of the pypa/cibuildwheel action. That is the reason, I think, that no wheels are built for Python 3.14.

```yaml
      - name: Build wheels
        uses: pypa/cibuildwheel@v2.23.3
```

Additionally, it is good security practice to refer to exact hashes of actions. You can then configure dependabot to make updates. If there is a desire, I can make a pull request for this.

Midnighter:
fix #1002

### DIFF
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -9,8 +9,8 @@ on:
     branches: [ master ]
 
 env:
-  latest_python: "3.13"
-  supported_pythons: '["3.9", "3.10", "3.11", "3.12", "3.13"]'
+  latest_python: "3.14"
+  supported_pythons: '["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]'
   miniforge_version: "23.11.0-0"
   miniforge_variant: "Mambaforge"
 
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -41,7 +41,7 @@ jobs:
           submodules: true
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.23.3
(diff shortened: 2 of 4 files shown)


wasade:
Thanks! It looks like there are some build failures?

Midnighter:
Yes, I was unable to judge whether they came because of the new version sinceithe error is on Python 3.11.

qiyunzhu:
The original test function is not deterministic. It isn't guaranteed that 100 random samples will definitely cover all combinations.

wasade:
Correct, good spot, rerunning failed jobs

Midnighter:
Not sure why the rerun could not be scheduled within 24 hours, but can you rerun it again, please?

wasade:
done

is macos-13 still valid?

Midnighter:
Actually not, 14 is the oldest https://github.com/actions/runner-images#available-images

I'll change that, too, when I get home.

wasade:
thanks!

Midnighter:
Made some changes to also test on Python 3.14. I'm not sure why this problem only pops up now with 3.14, but is there a good reason to install pytest via pip rather than a conda channel? It's also no longer recommended to use the conda defaults channel due to licensing issues.

wasade:
Good catch, any chance you could pull from conda-forge?

The gist is this project does not have direct funding, and as such I can only allocate minimal maintenance effort. We greatly appreciate contributions like yours which help immensely.

Midnighter:
Done already in the last commit 😉

wasade:
Thanks @Midnighter!

